PR #5, opened 2018-06-23
Viewing automatically created invoice

GSadee:
screencapture-localhost-8080-admin-invoices-dacf5b63-6c27-44f7-932c-c55b4324793a-2018-06-23-11_22_57

zrzut ekranu 2018-06-23 o 14 57 53


spec/Entity/BillingDataSpec.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    function it_implements_resource_interface(): void
+    {
+        $this->shouldImplement(ResourceInterface::class);
+    }
+
+    function it_has_proper_billing_data(): void
+    {
+        $this->beConstructedWith(
```
bartoszpietrzak1994: Since you don't need to change constructor parameters in several tests, why don't you put this in let() function?

spec/Entity/InvoiceSpec.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    function it_has_an_id(
+        BillingDataInterface $billingData,
+        LineItemInterface $lineItem
+    ): void {
+        $issuedAt = new \DateTimeImmutable('now');
+        $lineItems = new ArrayCollection([$lineItem->getWrappedObject()]);
+
+        $this->beConstructedWith(
```
bartoszpietrzak1994: Since you don't need to change constructor parameters in several tests, why don't you put this in let() function?

spec/Entity/LineItemSpec.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    function it_implements_resource_interface(): void
+    {
+        $this->shouldImplement(ResourceInterface::class);
+    }
+
+    function it_has_proper_line_item_data(): void
+    {
+        $this->beConstructedWith(
```
bartoszpietrzak1994: Since you don't need to change constructor parameters in serveral tests, why don't you put this in let() function?

spec/Event/OrderPlacedSpec.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    function it_represents_an_immutable_fact_that_an_order_has_been_placed(
+        OrderInterface $order
+    ): void {
+        $date = new \DateTimeImmutable('now');
+
+        $this->beConstructedWith($order, $date);
+
+        $this->order()->shouldBeLike($order);
+        $this->date()->shouldBeLike($date);
```
bartoszpietrzak1994: Considering PHP's DateTime precision, are you sure that this test will always pass?

tests/Behat/Page/Admin/ShowPage.php
```diff
@@ -20,10 +37,64 @@ public function getRouteName(): string
...
+    public function hasBillingData(
+        string $customerName,
+        string $street,
+        string $postcode,
+        string $city,
+        string $countryName
+    ): bool {
+        $billingDataText = $this->getElement('billing_address')->getText();
+
+        return
```
bartoszpietrzak1994: WDYT about 
```
stripos($billingDataText, $customerName) && stripos($billingDataText, $street) !== false && ...
```

?

spec/Entity/InvoiceSpec.php
GSadee: In this case, I need this function here to properly check not primitive parameters

spec/Event/OrderPlacedSpec.php
GSadee: I guess it should

tests/Behat/Page/Admin/ShowPage.php
GSadee: IMO it doesn't change anything, so I let it as it is

PR #12, opened 2018-07-03
I want to receive email with invoice when order is paid

bartoszpietrzak1994:
screen shot 2018-07-03 at 18 42 46

```
Scenario: I receive an email containing invoice once complete payment steps for order
    When the order "#00000666" is already paid
    Then an email containing invoice generated for order "#00000666" should be sent to "sylius@example.com"

Scenario: I don't receive an email containing invoice when the order is not paid
    When the order "#00000667" was cancelled
    Then an email containing invoice generated for order "#00000667" should not be sent to "sylius@example.com"
```

Scenarios fulfilled


src/EmailManager/InvoiceEmailManager.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\InvoicingPlugin\EmailManager;
+
+use Knp\Bundle\SnappyBundle\Snappy\Response\PdfResponse;
+use Sylius\Component\Mailer\Sender\SenderInterface;
+use Sylius\InvoicingPlugin\Entity\InvoiceInterface;
+
+final class InvoiceEmailManager implements InvoiceEmailManagerInterface
```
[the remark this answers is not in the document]
bartoszpietrzak1994: Since all similar classes in Sylius are `Managers` i wanted to stick to this pattern in order to be consistent. It's not the best name for it, however. 

It will be changed in review fixes commit

features/sending_invoice_when_order_is_paid.feature
```diff
@@ -0,0 +1,32 @@
+@managing_invoices
+Feature: Sending invoice when order is paid
+    In order to have a confirmation of order's payment
+    As a customer
```
Zales0123: `customer` -> `Customer`

features/sending_invoice_when_order_is_paid.feature
```diff
@@ -0,0 +1,32 @@
...
+        And I chose "Free" shipping method with "Cash on Delivery" payment
+        And I placed an order "#00000667"
+        And I bought a single "Angel T-Shirt"
+        And I addressed it to "Lucifer Morningstar", "Seaside Fwy", "90802" "Los Angeles" in the "United States"
+        And for the billing address of "Mazikeen Lilim" in the "Pacific Coast Hwy", "90806" "Los Angeles", "United States"
+        And I chose "Free" shipping method with "Cash on Delivery" payment
+
+    @ui
+    Scenario: I receive an email containing invoice once complete payment steps for order
+        When the order "#00000666" is already paid
```
Zales0123: `the order "#00000666" is already paid` seems more like *Given* step, rather than *When* :/ The best would be something like `When the "#00000666" has just been paid`, but it can be done in separate PR.

spec/Event/OrderPaymentPaidSpec.php
```diff
@@ -0,0 +1,21 @@
...
+final class OrderPaymentPaidSpec extends ObjectBehavior
+{
+    function it_represents_an_immutable_fact_that_payment_related_to_order_was_completed(): void
+    {
+        $date = new \DateTimeImmutable('now');
+        $orderNumber = '000001';
+
+        $this->beConstructedWith($orderNumber, $date);
+
+        $this->orderNumber()->shouldBeEqualTo($orderNumber);
```
Zales0123: Why `shouldBeEqualTo` not `shouldReturn`?

Zales0123: And I think it's not a problem to use just a value here and above
```
$this->beConstructedWith('000001', $date);

$this->orderNumber()->shouldReturn('000001');
```

src/Event/OrderPaymentPaid.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\InvoicingPlugin\Event;
+
+use DateTime;
```
Zales0123: Not needed

src/Event/OrderPaymentPaid.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public function orderNumber(): string
+    {
+        return $this->payload()['order_number'];
+    }
+
+    public function date(): \DateTimeInterface
+    {
+        return DateTime::createFromFormat('d-m-Y H:i:s.u', $this->payload()['date']);
```
Zales0123: `\DateTime`

src/EventListener/OrderPaymentPaidListener.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->orderRepository = $orderRepository;
+        $this->templatingEngine = $templatingEngine;
+    }
+
+    public function __invoke(OrderPaymentPaid $event)
+    {
+        /** @var InvoiceInterface $invoice */
+        $invoice = $this->invoiceRepository->findOneByOrderNumber($event->orderNumber());
+
+        $filename = sys_get_temp_dir() . '/' . sprintf('invoice-%s.pdf', $invoice->id());
```
Zales0123: Could we extract this to private function with explainable? It just looks awful near object-oriented code :)

src/Email/InvoiceEmailSender.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\InvoicingPlugin\Email;
+
+use Sylius\Component\Mailer\Sender\SenderInterface;
+use Sylius\InvoicingPlugin\Entity\InvoiceInterface;
+
+final class InvoiceEmailSender implements InvoiceEmailManagerInterface
```
Zales0123: `InvoiceEmailManagerInterface` -> `InvoiceEmailSenderInterface`

src/Repository/DoctrineInvoiceRepository.php
```diff
@@ -36,4 +36,12 @@ public function add(InvoiceInterface $invoice): void
         $this->entityManager->persist($invoice);
         $this->entityManager->flush();
     }
+
+    public function findOneByOrderNumber(string $orderNumber): InvoiceInterface
+    {
+        /** @var InvoiceInterface $invoice */
+        $invoice = $this->entityRepository->findOneBy(['orderNumber' => $orderNumber]);
+
+        return $invoice;
```
Zales0123: Maybe just `return $this->entityRepository->findOneBy(['orderNumber' => $orderNumber]);`?

tests/Behat/Context/Domain/InvoiceEmailContext.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->emailChecker = $emailChecker;
+    }
+
+    /**
+     * @Then an email containing invoice generated for order :orderNumber should be sent to :recipient
+     */
+    public function emailContainingInvoiceForOrderShouldBeSent(string $orderNumber, string $recipient): void
+    {
+        Assert::true($this->emailChecker->hasMessageTo(sprintf('was generated for order with number %s', $orderNumber), $recipient));
```
Zales0123: We could move `sprintf` to next line to make this one a little bit shorter

src/Repository/DoctrineInvoiceRepository.php
bartoszpietrzak1994: Unfortunately, type hint placed in the comment is needed in order to make this code pass phpstan checks and 

```
/** @var InvoiceInterface $invoice */
return $this->entityRepository->findOneBy(['orderNumber' => $orderNumber]);
```

doesn't seem to work.

src/DependencyInjection/SyliusInvoicingExtension.php
```diff
@@ -48,7 +48,17 @@ private function prependWinzouStateMachine(ContainerBuilder $container): void
...
                 ],
             ],
-        ]);
+            'sylius_payment' => [
+                'callbacks' => [
+                    'after' => [
+                        'sylius_invoicing_plugin_payment_complete_producer' => [
+                            'on' => ['complete'],
+                            'do' => ['@Sylius\InvoicingPlugin\EventListener\OrderPaymentPaidProducer', '__invoke'],
+                            'args' => ['object'],
```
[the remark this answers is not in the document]
pamil: This saves us another installation step, being "Add import to your `config.yml`".

src/Repository/DoctrineInvoiceRepository.php
pamil: I think we need to use https://github.com/phpstan/phpstan-webmozart-assert and `Assert::notNull($invoice)` / `Assert::isInstanceOf($invoice, InvoiceInterface::class)` between setting the variable and return.

src/Repository/InvoiceRepository.php
```diff
@@ -10,5 +10,7 @@
 {
     public function get(string $invoiceId): InvoiceInterface;
 
+    public function findOneByOrderNumber(string $orderNumber): InvoiceInterface;
```
pamil: There's a convention for repository methods:
 - `get*` - when it returns an entity or throws an exception
 - `find*` - when it returns an entity or null

The current implementation is more `get*`-like.

src/Resources/views/Invoice/Download/pdf.html.twig
```diff
@@ -42,19 +40,19 @@
                         {{ item.variantCode }}
                     </td>
                     <td>
-                        {{ money.format(item.unitPrice, invoice.currencyCode) }}
+                        {{ item.unitPrice|sylius_format_money(invoice.currencyCode, sylius.localeCode) }}
```
pamil: Could we also save `localeCode` on an invoice? We can take it directly from an order instance.

src/EventListener/OrderPaymentPaidListener.php
```diff
@@ -0,0 +1,82 @@
...
+        /** @var InvoiceInterface $invoice */
+        $invoice = $this->invoiceRepository->findOneByOrderNumber($event->orderNumber());
+
+        $filePath = $this->generateTemporaryPdfFilePathBasedOnInvoiceId($invoice->id());
+
+        $this->pdfGenerator->generateFromHtml(
+            $this->templatingEngine->render('@SyliusInvoicingPlugin/Resources/views/Invoice/Download/pdf.html.twig', [
+                'invoice' => $invoice
+            ]),
+            $filePath
```
pamil: It looks kinda weird right now to have this file created implicitly (`$this->pdfGenerator->generateFromHtml(...)`) and then removing it explicitly (`$this->removeTemporaryPdfFile(...);`).

My suggestion would be to extract file-related operations from this class, making it something like:

```php
$invoiceContent = $this->pdfGenerator->generateOutputFromHtml(...);

$this->temporaryFilesystem->create($file, $invoiceContent);

try {
  $this->emailSender->sendInvoiceEmail();
} finally {
  $this->temporaryFilesystem->remove($file);
}
```

Usage of that `finally` block will make sure that even if sending that email fails, the invoice gets removed afterwards.

As for the future, I'd think about extracting this whole listener logic to a command handler (and just dispatching SendInvoiceEmail command), so that it will be possible to handle it asynchronously and the logic will be separated (listening on event / reacting on event).

src/DependencyInjection/SyliusInvoicingExtension.php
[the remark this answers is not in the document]
pamil: If the question is more about PHP / YAML choice than prepending the configuration, I think we can parse a YAML file and then prepend its contents, but I'd need to check that first.

src/Resources/views/Invoice/Download/pdf.html.twig
bartoszpietrzak1994: I think that we could. However, it would create an issue and do it in a separate PR.

pamil: #13

src/EventListener/OrderPaymentPaidListener.php
pamil: Btw. this class is missing a unit test :)

features/sending_invoice_when_order_is_paid.feature
```diff
@@ -0,0 +1,32 @@
...
+        And for the billing address of "Mazikeen Lilim" in the "Pacific Coast Hwy", "90806" "Los Angeles", "United States"
+        And I chose "Free" shipping method with "Cash on Delivery" payment
+
+    @ui
+    Scenario: I receive an email containing invoice once complete payment steps for order
+        When the order "#00000666" has just been paid
+        Then an email containing invoice generated for order "#00000666" should be sent to "sylius@example.com"
+
+    @ui
+    Scenario: I don't receive an email containing invoice when the order is not paid
```
Zales0123: Should be `Not receiving an email containing invoice when the order is not paid` (similar above), but it could be fixed in next PR :)

src/Event/OrderPaymentPaid.php
```diff
@@ -0,0 +1,30 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\InvoicingPlugin\Event;
+
+use Prooph\Common\Messaging\DomainEvent;
+use Prooph\Common\Messaging\PayloadTrait;
+
+/** @final */
+class OrderPaymentPaid extends DomainEvent
```
Zales0123: Why it's not final?

Zales0123:
Thank you Bartosz! 🦅

PR #24, opened 2018-07-12
Improved readme and missing config file

bartoszpietrzak1994:


tests/Behat/Resources/services.xml
```diff
@@ -3,6 +3,7 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <parameters>
         <parameter key="sylius.behat.page.admin.channel.update.class">Tests\Sylius\InvoicingPlugin\Behat\Page\Admin\Channel\UpdatePage</parameter>
+        <parameter key="sylius_invoicing_plugin.knp_snappy.pdf.binary">/usr/local/bin/wkhtmltopdf</parameter>
```
Zales0123: This parameter should be also configured for dev env, without that we would **force** user to configure it rather than providing some default configuration. We can configure it in `src/Resources/config/config.yml` under `parameters:` key. Then it can be mentioned in README that there is a default value set but it can be customized

src/Resources/config/config.yml
```diff
@@ -0,0 +1,11 @@
+sylius_mailer:
+    emails:
+        invoice_generated:
+            subject: sylius.emails.invoice_generated.subject
+            template: "@SyliusInvoicingPlugin/Resources/views/Invoice/Email/invoiceGenerated.html.twig"
+
+knp_snappy:
+    pdf:
+        enabled: true
+        binary: %sylius_invoicing_plugin.knp_snappy.pdf.binary%
```
pamil: To be honest, I don't think we need a parameter here, as anyone can overwrite this configuration by adding the following snippet to their config file:

```yaml
knp_snappy:
    pdf:
        binary: foobar
```

pamil: What's more, if there are two plugins configuring this `knp_snappy`'s pdf binary and this one is the first one, then it would have no effect.

Zales0123:
🎉

PR #35, opened 2018-07-31
Fixed currency symbols rendering with UTF-8 encoding

mamazu:
Some currency symbols are rendered weirdly if you don't specify the charset in the rendered HTML page. (eg. the € symbol)


pamil:
Thank you, Maximilian! 🥇

PR #36, opened 2018-07-31
Resetting sequence index every month

### LINKED ISSUE #20: Sequence number placed in invoiceNumber should be reset every month

GSadee:
Fixes #20

### DIFF
--- a/src/Generator/SequentialInvoiceNumberGenerator.php
+++ b/src/Generator/SequentialInvoiceNumberGenerator.php
@@ -49,15 +49,22 @@ final class SequentialInvoiceNumberGenerator implements InvoiceNumberGenerator
 
     public function generate(): string
     {
-        $invoiceIdentifierPrefix = $this->dateTimeProvider->__invoke()->format('Y/m') . '/';
+        $currentDate = $this->dateTimeProvider->__invoke();
+
+        $invoiceIdentifierPrefix = $currentDate->format('Y/m') . '/';
 
         /** @var InvoiceSequenceInterface $sequence */
         $sequence = $this->getSequence();
 
         $this->sequenceManager->lock($sequence, LockMode::OPTIMISTIC, $sequence->getVersion());
 
+        if ($this->isMonthChanged($currentDate, $sequence->getLastGeneratedAt())) {
+            $sequence->resetIndex();
+        }
+
         $number = $this->generateNumber($sequence->getIndex());
         $sequence->incrementIndex();
+        $sequence->setLastGeneratedAt($currentDate);
 
         return $invoiceIdentifierPrefix . $number;
     }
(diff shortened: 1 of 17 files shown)


src/Generator/SequentialInvoiceNumberGenerator.php
```diff
@@ -49,15 +49,22 @@ public function __construct(
...
+
+        $invoiceIdentifierPrefix = $currentDate->format('Y/m') . '/';
 
         /** @var InvoiceSequenceInterface $sequence */
         $sequence = $this->getSequence();
 
         $this->sequenceManager->lock($sequence, LockMode::OPTIMISTIC, $sequence->getVersion());
 
+        if ($this->isMonthChanged($currentDate, $sequence->getLastGeneratedAt())) {
+            $sequence->resetIndex();
```
pamil: This might cause issues when events are processed asynchronously and not in chronological order, eg. you've got OrderPlaced at 31.03, OrderPlaced at 01.04 and OrderPlaced at 31.03 again.

The solution I'd see for that is to replace `lastGeneratedAt` with `group` (which will be `Y/m`) and always get the sequence for a given group. It removes the need for this conditional and resetting index also.

src/Generator/SequentialInvoiceNumberGenerator.php
```diff
@@ -49,15 +49,22 @@ public function __construct(
...
+        }
+
         $number = $this->generateNumber($sequence->getIndex());
         $sequence->incrementIndex();
```
pamil: Not connected with this PR, but could use a small improvement - there's no other use case than to call `getIndex()` and `incrementIndex()`  together. What about merging those two into one method `getNextIndex()`?

src/Generator/SequentialInvoiceNumberGenerator.php
```diff
@@ -49,15 +49,22 @@ public function __construct(
 
     public function generate(): string
     {
-        $invoiceIdentifierPrefix = $this->dateTimeProvider->__invoke()->format('Y/m') . '/';
+        $currentDate = $this->dateTimeProvider->__invoke();
```
pamil: This can also cause some issues, as you probably want the invoice number to be generated just when the order is placed / order payment is paid, so there's a rare edge case when the month changes between those. What do you think about modifying `InvoiceNumberGeneratorInterface` to accept `\DateTimeInterface` as an argument in `generate()` method, so that we won't need to rely on the current server time?

tests/Behat/Context/Setup/InvoiceContext.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function theLastInvoiceHasBeenGeneratedToday(int $number): void
+    {
+        $sequence = $this->getSequence();
+        $sequence->setLastGeneratedAt(new \DateTimeImmutable('now'));
+        for ($i = 0; $i < $number; ++$i) {
+            $sequence->incrementIndex();
+        }
+
+        $this->sequenceManager->flush();
```
pamil: These steps implementation looks kinda hacky, as it's strictly connected to the invoice number sequence implementation, while the step itself only mentions invoices - creating test orders (and then invoices based on that) would be a more comprehensive solution.

Issue #26, opened 2018-07-12
Missing "back" button on invoice show page


Zales0123:
<img width="1172" alt="zrzut ekranu 2018-07-12 o 14 16 08" src="https://user-images.githubusercontent.com/6212718/42632681-4591b9e4-85de-11e8-845f-4ab7216b1bee.png">

GSadee:
But why? User can just click the back button in the browser 🎉 We don't have these buttons anywhere else. /cc @CoderMaggie

Zales0123:
For me personally limiting requirements of "back" browser button usages is always a good UX :)

Roshyo:
I do agree with @Zales0123 
Plus, back buttons in the browser are not always used by Final users (I mean the shop administrators)

CoderMaggie:
Let's just add a back button, but no text, simply an icon 🔙 :)

PR #38, opened 2018-07-31
Back button on invoice show page

bartoszpietrzak1994:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/26


src/Resources/views/Invoice/show.html.twig
```diff
@@ -12,8 +12,10 @@
             {% include '@SyliusInvoicingPlugin/Invoice/Show/_header.html.twig' %}
         </div>
         <div class="six wide right aligned column">
-            {% set path = path('sylius_invoicing_plugin_admin_invoice_download', { 'id': invoice.id }) %}
+            {% set path = path('sylius_invoicing_plugin_admin_invoice_back', { 'id': invoice.id }) %}
```
pamil: Why not `path('sylius_invoicing_plugin_admin_invoice_index')`?

src/Resources/config/app/routing/admin_invoicing.yml
```diff
@@ -45,3 +45,11 @@ sylius_invoicing_plugin_admin_order_invoices_partial:
             repository:
                 method: findByOrderNumber
                 arguments: $orderNumber
+
+sylius_invoicing_plugin_admin_invoice_back:
```
pamil: Why don't we just use the invoice index page route directly?

tests/Behat/Context/Ui/Admin/ManagingInvoicesContext.php
```diff
@@ -209,4 +209,22 @@ public function downloadInvoice(): void
...
+    {
+        $this->showPage->goBack();
+    }
+
+    /**
+     * @Then I should see all invoices
+     */
+    public function shouldSeeAllInvoices(): void
+    {
+        $this->indexPage->waitForRedirect(3);
```
pamil: It shouldn't be needed (unless it's run in Selenium).

Issue #13, opened 2018-07-05
Add locale code to the invoice model


pamil:
See https://github.com/Sylius/InvoicingPlugin/pull/12#discussion_r200174216.

PR #39, opened 2018-08-01
Locale code added to invoice model

bartoszpietrzak1994:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/13


migrations/Version20180801083047.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
```
pamil: CS, could be fixed automatically.

pamil:
Thank you, Bart! 🎉

PR #40, opened 2018-08-02
Make config great again

pamil:


teohhanhui:
This is really bad... :(

teohhanhui:
It's something that Sylius needs to stop doing for us to get out of config "BC" hell...

PR #45, opened 2018-08-22
Bring back support for stable version of Sylius

teohhanhui:


Zales0123:
I think that these plugins were targeting to base on Sylius `1.3` from their very beginning... Do we really want to support earlier Sylius versions? cc @pamil @bartoszpietrzak1994

pamil:
The argument against supporting Sylius 1.2 here is our inability to test the plugin with that version of Sylius, as there is only one test app inside this plugin and it works with Sylius 1.3+ (mostly due to JS-related changes made).

teohhanhui:
Indeed, but I'd argue that barring our ability to test it, we should just support it on a "best effort" basis. It could be clearly documented...

PR #47, opened 2018-08-26
Making build great again

bartoszpietrzak1994:


pamil:
Fixing Sylius/Sylius related issues now.

pamil:
Should be passing after Sylius/Sylius#9676 is merged.

.gitignore
```diff
@@ -9,3 +9,5 @@
 !/etc/build/.gitkeep
 
 /tests/Application/yarn.lock
+
+.env.test
```
pamil: `/.env.test` is a good practice, as right now it would ignore `.env.test` files even in subdirectories (which is something not really expected here).

composer.json
```diff
@@ -11,7 +11,8 @@
         "prooph/service-bus": "^6.2",
         "prooph/service-bus-symfony-bundle": "^0.7.0",
         "ramsey/uuid": "^3.7",
-        "knplabs/knp-snappy-bundle": "^1.5"
+        "knplabs/knp-snappy-bundle": "^1.5",
+        "friends-of-behat/page-object-extension": "^0.2.0"
```
pamil: It is a dev dependency, so should go to `require-dev`.

pamil:
Fixed the PR for you 💃

PR #48, opened 2018-08-28
Remove hardcoded binary path for wkhtmltopdf

teohhanhui:
KnpSnappyBundle already provides a more sensible default: https://github.com/KnpLabs/KnpSnappyBundle/blob/v1.5.2/DependencyInjection/Configuration.php#L41


pamil:
Thanks, Teoh! 🥇

PR #65, opened 2018-10-02
Add messages for fr locale

### LINKED ISSUE #44: Missing French translation 🥖
Hi,

Is that possible to have this plugin on crowdin please?
I would like to translate it 😃 .

Thank you!

alanpoulain:
Fix #44

There is an issue with the colon in Sylius. In French we need to have a **non-breaking space** before it. That's why it should be in the message itself and not in the code.
An example:
https://github.com/Sylius/InvoicingPlugin/blob/e91c56d357736df712ee0980261a914c04fc38a4/src/Resources/views/Invoice/Download/pdf.html.twig#L10
WDYT?

### DIFF
--- a/src/Resources/translations/messages.fr.yml
+++ b/src/Resources/translations/messages.fr.yml
@@ -1,0 +1,17 @@
+sylius_invoicing_plugin:
+    ui:
+        billing_data: 'Données de facturation'
+        buyer: 'Acheteur'
+        download_invoice: 'Télécharger'
+        invoice: 'Facture'
+        invoice_id: 'ID de facture'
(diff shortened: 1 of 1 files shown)


alanpoulain:
Yes, I had the same idea in mind :slightly_smiling_face:
To comfort the issue, here is an interesting answer: https://softwareengineering.stackexchange.com/questions/234546/colons-in-internationalized-ui

Zales0123:
Ok, so I'm merging this one, thank you, Alan, for your help and feel free to open another issue with the problem explanation :) 🎉

PR #67, opened 2018-10-05
Describe installation with Symfony Flex

Zales0123:
**17 additions & 59 deletions** in `README` file, Symfony Flex totally rocks 😄 🤘

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,57 +30,25 @@ Additional feature of the plugin that fulfills Invoicing domain is the ability t
     ```bash
     composer require sylius/invoicing-plugin
     ```
-
-2. Import configuration:
-
-    ```yaml
-    imports:
-        - { resource: "@SyliusInvoicingPlugin/Resources/config/config.yml" }
-    ```
-
-3. Import routing:
-
-    ```yaml
-    sylius_invoicing_plugin_admin:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,57 +30,25 @@ Additional feature of the plugin that fulfills Invoicing domain is the ability t
...
-        new \Knp\Bundle\SnappyBundle\KnpSnappyBundle(),
-        new \Prooph\Bundle\ServiceBus\ProophServiceBusBundle(),
-        new \Sylius\InvoicingPlugin\SyliusInvoicingPlugin(),
-    ];
+    ```bash
+    cp -R vendor/sylius/invoicing-plugin/migrations/* src/Migrations
+    bin/console doctrine:migrations:migrate
     ```
 
-5. Check if you have `wkhtmltopdf` binary. If not, you can download it [here](https://wkhtmltopdf.org/downloads.html).
```
bartoszpietrzak1994: Shouldn't we still inform that such binary is needed to make InvoicingPlugin work properly and it's path is placed in `.env` file?

Zales0123: That was also my intention... But I'm not 100% sure as well 😄

bartoszpietrzak1994: Right, I forgot that knp-snappy-bundle is also powered by Symfony Flex :D

PR #76, opened 2018-10-30
Implement Symfony's Voter to determine if user is able to access an invoice

pamil:


spec/Security/Voter/InvoiceVoterSpec.php
```diff
@@ -0,0 +1,91 @@
...
+        $token->getUser()->willReturn($shopUser);
+
+        $shopUser->getCustomer()->willReturn($customer);
+
+        $orderRepository->findOneByNumberAndCustomer('1337', $customer)->willReturn(null);
+
+        $this->vote($token, $invoice, [InvoiceVoter::ACCESS])->shouldReturn(VoterInterface::ACCESS_DENIED);
+    }
+
+    function it_allows_accessing_an_invoice_if_user_has_made_the_order_connected_to_the_invoice(
```
bartoszpietrzak1994: `it_does_not_allow_accessing_an_invoice_if_user_has_placed_the_order_related_to_the_invoice ` ?

src/Security/Voter/InvoiceVoter.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    protected function voteOnAttribute($attribute, $subject, TokenInterface $token): bool
+    {
+        /** @var InvoiceInterface $subject */
+        Assert::isInstanceOf($subject, InvoiceInterface::class);
+
+        $user = $token->getUser();
+
+        if (!$user instanceof UserInterface) {
```
bartoszpietrzak1994: I'm not sure about this comment, but shouldn't it be `if (!($user instanceof UserInterface))` instead of `if (!$user instanceof UserInterface)` ?

src/Security/Voter/InvoiceVoter.php
```diff
@@ -0,0 +1,74 @@
...
+            return false;
+        }
+
+        switch ($attribute) {
```
bartoszpietrzak1994: Yay, a `switch`! Haven't seen that in a long time 🎉

src/Security/Voter/InvoiceVoter.php
pamil: `instanceof` has a stronger precedence than `!`.

src/Security/Voter/InvoiceVoter.php
lchrusciel: :(

lchrusciel: I'm not sure why don't we just used an old fashion early return :(

PR #77, opened 2018-10-30
Make emails sending more straightforward

pamil:
Based on #76.

PR #78, opened 2018-10-30
Switch to Symfony Messenger

pamil:


Zales0123:
I suppose here we have a similar problem as in https://github.com/Sylius/RefundPlugin/pull/89. If we want to use Symfony messenger, I think we need to allow only `^4.1` symfony libraries... which does not allow to use this plugin by Sylius application that still bases on Symfony 3.4 😕 do you see any solution of this problem?

pamil:
I tried to isolate these dependencies but failed. I'd go with Symfony 4.1+ then, Symfony is stable enough not to rely only on LTS versions.

Zales0123:
Tried the same and also failed (it's not only me 🎉 😄). Supporting only **Symfony 4.1+** seems fine for me (will do the same for `RefundPlugin`). cc @Sylius/core-team

bartoszpietrzak1994:
Introducing Symfony Messenger is a welcome improvement. However, since it requires to abandon Symfony 3.4 support I'd consider making this change as transparent as possible or even releasing two versions of InvoicingPlugin for a short period of time.

What's more, maybe we should start abandoning Symfony 3.4 support by disabling it in `Sylius-Standard` project? Starting this whole process in plugins' `composer.json` files may result in low DX experience IMO.

pamil:
I wouldn't go as far as deprecating it in Sylius core now (waiting for the community survey answers), but I think it's fine for now to do so in plugins. I'd release InvoicingPlugin after #76 and #77 are merged, and then a new minor after this PR is merged (and put some info about it in release notes).

bartoszpietrzak1994:
@pamil Now that's a nice approach 👍

bartoszpietrzak1994:
It's ok for me, however, I'd like @Zales0123 to provide another approve for this PR since he's already into Symfony Messenger component 🎉

src/Resources/config/config.yml
```diff
@@ -8,11 +8,14 @@ knp_snappy:
...
         enabled: true
 
-prooph_service_bus:
-    event_buses:
-        sylius_invoicing_event_bus: ~
-    command_buses:
-        sylius_invoicing_command_bus: ~
+framework:
+    messenger:
+        default_bus: sylius_invoicing_plugin.command_bus
```
Zales0123: I would extract it to test configuration file as [it has been done in RefundPlugin](https://github.com/Sylius/RefundPlugin/pull/89/files#diff-3cb4d392b9fb5e2d3cf5e66c4b081c7fR3)

PR #87, opened 2018-11-27
Resolving suites/tags conflict

bartoszpietrzak1994:
It occurred that there already is a `@managing_channels` tag in Sylius/Sylius as well as `ui_managing_channels` suite. Therefore, channel-related tests from InvoicingPlugin were loaded by suite used by Sylius/Sylius and fail due to undefined steps. 

The problem occurred only while running InvocingPlugin tests with Sylius/Sylius installed as a dependency.


Zales0123:
Thank you, Bartosz! :tada:

PR #101, opened 2019-01-29
ShopBillingData id should be int

pierre-H:
Like #100


bartoszpietrzak1994:
Hi Pierre!

Thank you for your contribution. We've decided to move ShopBillingData to sylius-core in version 1.4 (and also change the id type from string to int). The pull request containing the update to Sylius in version 1.4 is open right now and will probably be merged soon. Hence, the ShopBillingData entity will disappear from InvoicingPlugin.

Since the InvoicingPlugin has not been released in a stable version yet, the only way to use its new features will require updating Sylius version.

pierre-H:
Ok, thank you !
So I close this PR

PR #109, opened 2019-02-17
Fix Channel form

Zales0123:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/105

Issue #84, opened 2018-11-19
customize template


itwhy:
Hi, i tried to customise pdf template like other bundle but nothing change : / 

my path is Templates/SyliusInvoicingPlugin/Invoice/Download/ ..

edit : the path is actually : Templates/bundles/SyliusInvoicingPlugin/Invoice/Download/ ..

do you know why ? 

thx you !

Zales0123:
Hello Mathieu, in fact, it should work without a problem :/ I'll try to investigate it asap and give you some feedback.

Zales0123:
I've just checked it and indeed there is some problem with overriding 😢 Maybe it's because plugin templates are placed in `Resources/views`? Thank you for the report, anyway, we will try to find a solution as soon as possible. Of course, if you have any idea what can be wrong, feel free to contribute :)

teohhanhui:
Could it have anything to do with Symfony's bundle naming convention?

itwhy:
haha, i'd really like to help ! but i'm just a "user", and very sylius noob :)

itwhy:
hi guys, any news for this issue ?

bartoszpietrzak1994:
Hello Mathieu,

Unfortunately, there's a bug in Sylius due to a conflict between new Twig and Sylius' ThemeBundle conventions since 1.3.1 version. Although we are working on a solution, right now I'd suggest downgrading Sylius to 1.3.0 version if you are in a hurry.

nemoneph:
Hello,
I'm facing the same problem on 1.4-BETA.

Any news on this ?
Thx :)

bartoszpietrzak1994:
Hello @itwhy @nemoneph !

We've managed to dig into this problem and managed to find a solution that allows us to customize templates under `templates/Bundles/SyliusInvoicingPlugin` path. 

The new version of SyliusInvoicingPlugin has just been relased (`v0.8.3`) so please update it via composer and let us know if everything works like a charm :)

itwhy:
hi guys, 
sorry i found another way to do it (app/Ressources/SyliusInvoicingPlugin/) and i'm now too late to test the new way, the deadline is too close.
i'll try that in the next v.
anyway, thx you ! hope this will help others guys ^^

PR #110, opened 2019-02-18
Make invoice beautiful again

Zales0123:
Based on https://github.com/Sylius/InvoicingPlugin/pull/109

Even though I know we should not focus on providing a SUPER AMAZINGLY BEAUTIFUL invoice, it's good to have it at least acceptably nice... And if it can be customized just by changing the logo, it's even better :) The design is based on https://github.com/sparksuite/simple-html-invoice-template.

Before: 

zrzut ekranu 2019-02-18 o 10 03 46

After:

zrzut ekranu 2019-02-18 o 10 00 52


src/Resources/views/Invoice/Download/pdf.html.twig
```diff
@@ -1,125 +1,109 @@
...
-        <br /><br />
+<body>
+<div class="invoice-box">
+    <table cellpadding="0" cellspacing="0">
+        <tr class="top">
+            <td colspan="4">
+                <table>
+                    <tr>
+                        <td colspan="3" class="title">
+                            <img src="https://sylius.com/assets/sylius-logo.png" style="width:300px;">
```
bartoszpietrzak1994: WDYT about downloading the picture and including it into the project? I'm not sure if we should rely on an external resource in such case like generating an invoice.

src/Resources/config/config.yml
```diff
@@ -8,6 +8,10 @@ knp_snappy:
     pdf:
         enabled: true
 
+twig:
+    globals:
+        invoiceLogoPath: "%kernel.project_dir%/vendor/sylius/invoicing-plugin/assets/sylius-logo.png"
```
pamil: How can we make sure that it works? Why do we need another path for the test environment?

Zales0123: Because in a plugin test application we don't have an invoice plugin in vendors and in a real application we do :/ I'm not pleased with that approach, but I don't see any other way to force KnpSnappyBundle to render the image that is inside the repository. 

The ideal scenario would be to check the PDF content in a Behat test - it's possible, afaik, but requires some work, definitely not for this PR. For now, I've just checked the new invoice template in a Sylius-Standard application and it worked with no problems 🚜

pamil: Can we pass a relative path using a bundle notation? `@SyliusInvoicingPlugin/logo.png` and then use `Symfony\Component\HttpKernel\Config\FileLocator` to resolve it?

Zales0123: I've tried it and it didn't work :| I see the images problem is one of the most common: https://github.com/KnpLabs/KnpSnappyBundle/issues/82, https://github.com/KnpLabs/KnpSnappyBundle/issues/232, https://github.com/KnpLabs/KnpSnappyBundle/issues/204, https://github.com/KnpLabs/snappy/blob/master/doc/faq.md#q-how-to-convert-page-with-relative-links-from-stdin--how-to-use-relative-media-urls

Zales0123: Ok, apparently it works xD thanks @pamil

Zales0123:
Ok, right now we have a logo path and template path both injected into the `InvoicePdfFileGenerator`. It's not the best solution but allows to override both of them (by overriding the service and passing custom path). I would vote to merge it as it is, as we still need to work on invoice overwrite - see https://github.com/Sylius/InvoicingPlugin/issues/84 🐃

bartoszpietrzak1994:
Thank you, Mateusz! :tada:

PR #116, opened 2019-03-13
Make all plugin templates possible to overwrite

Zales0123:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/84

Ok, I think we've finally managed (kinda) to resolve this problem 😄 There are some problems in Sylius (or even in Symfony itself) with overwriting templates, that are not configured with a one, specific path style. It should always follow `@SyliusTestPlugin/Catalog/templateName.html.twig` convention - and as you can see, there were some other conventions used :) It would be good if someone could test these changes before merging (cc @itwhy), but I've tested all plugin templates on a fresh Sylius-Standard application and it worked 🚀


bartoszpietrzak1994:
Thanks, Mateusz! :tada:

PR #102, opened 2019-02-07
Update sylius version to 1.4

bartoszpietrzak1994:

PR #123, opened 2019-04-18
Missing migration file

bartoszpietrzak1994:
It took a (long) while to discover this issue but it seems that we've missed that a proper migration file should have been attached to https://github.com/Sylius/InvoicingPlugin/pull/102 🏅

### DIFF
--- a/migrations/Version20190418115432.php
+++ b/migrations/Version20190418115432.php
@@ -12,7 +12,7 @@ use Doctrine\Migrations\AbstractMigration;
  */
 final class Version20190418115432 extends AbstractMigration
 {
-    public function up(Schema $schema) : void
+    public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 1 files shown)


migrations/Version20190418115432.php
```diff
@@ -0,0 +1,30 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190418115432 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
```
GSadee: Unnecessary space before `:`

bartoszpietrzak1994:
After a while I realized that this PR is kind of dummy - I forgot that we still store shop billing data on every invoice in order to store an immutable set of data once an administrator decides to change channel's shop billing data, to be closed

Issue #117, opened 2019-03-13
Make invoices and credit memos grid consistent


Zales0123:
Invoices and credit memos (from [RefundPlugin](https://github.com/Sylius/RefundPlugin)) are twin concepts and therefore, I believe, we should make them consistent as much as it's possible.

Looking at credit memos and invoices grids on order show page:

<img width="874" alt="Zrzut ekranu 2019-03-13 o 12 50 48" src="https://user-images.githubusercontent.com/6212718/54277056-1109b700-458f-11e9-83a0-c8dcfd8ad629.png">

And on overall indexes: 

<img width="1138" alt="Zrzut ekranu 2019-03-13 o 12 55 41" src="https://user-images.githubusercontent.com/6212718/54280180-2a166600-4597-11e9-9098-1562d45fb78a.png">
<img width="1139" alt="Zrzut ekranu 2019-03-13 o 12 55 46" src="https://user-images.githubusercontent.com/6212718/54280182-2aaefc80-4597-11e9-8852-4c5a8b2d584c.png">

we can see a lot of things that can be unified 🤖 

## Possible solution

My proposition for both invoices and credit memos overall grids is something like this: 

<img width="1168" alt="Zrzut ekranu 2019-03-13 o 14 04 59" src="https://user-images.githubusercontent.com/6212718/54281001-1c61e000-4599-11e9-9a81-98f9b7585fec.png">

1. Invoice/credit memo number
2. Order number with `#` at the beginning and a link to order show page
3. Channel name with a colorful circle
4. Total
5. Issued at date, in `Y-m-d` format
6. Buttons
    - "Show" - green, eye
    - "Download" - purple, download icon
    - "Resend" - yellow, paper plane (not needed for credit memo [for now](https://github.com/Sylius/RefundPlugin/issues/122))

On partial grids (displayed on order show page) we can get rid of 2. and 3. column, as they're irrelevant.

Tell me what you think about this proposition 🖖 I feel strongly that these two plugins can be stabilized in the very nearest future 🚀 🔥

CoderMaggie:
I love the plan to unify these.

Although I vote for the buttons to be _greyish_ instead of colorful :)

PR #124, opened 2019-04-18
Make invoices and credit memo grid consistent

bartoszpietrzak1994:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/117

Screenshot 2019-04-18 at 16 36 59


src/Provider/ChannelColorProviderInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\InvoicingPlugin\Provider;
+
+interface ChannelColorProviderInterface
+{
+    public function __invoke(string $channelCode): string;
```
lchrusciel: It is just my personal opinion, and I'm not sure about standards on this plugin, but I found it cleaner to name functions if they are part of the interface. IMHO `__invoke` is perfect but for controllers, command handlers and this type of classes (without interfaces). Even simple:

```suggestion
    public function provide(string $channelCode): string;
```

bartoszpietrzak1994: I get your point but what can a *Provider do instead of... providing? I can see both pros and cons of using __invoke() but to me, channelColorProvider->provide() sounds a bit redundant. WDYT?

lchrusciel: That's correct, and maybe the problem is just with my aesthetic feeling or that I'm used to these redundant functions. On the other hand, `__invoke` is a magic method and is meant to be used like: `$class($arg)`. When we are using it with `$class->__invoke($arg)` is the same as we would agree for example for `run` function. It does not provide any benefit and looks bad for me. 

IMHO lines like https://github.com/Sylius/InvoicingPlugin/pull/124/files#diff-94b24b9abcb05ca0be18c811d5d20a32R24 is a trigger for wrong `__invoke` usage. In such a case I prefer redundant functions, so at least we don't have a call for a function with two dashes at the beginning.

README.md
```diff
@@ -118,6 +118,12 @@ Sonata events. You can read about customizing templates via events here:
 
 <http://docs.sylius.com/en/latest/customization/template.html>
 
+Invoicing Plugin renders invoices grid using a certain pattern, including displaying a dot next to channel in which an invoice as issued.
+The dot's color is dependant on a property defined on Channel entity or, if not provided, a global parameter named `default_channel_color`.
+
+Like any 
+ 
+
```
GSadee: Unnecessary blank line

src/Provider/ChannelColorProvider.php
```diff
@@ -0,0 +1,31 @@
...
+        $this->channelRepository = $channelRepository;
+        $this->defaultChannelColor = $defaultChannelColor;
+    }
+
+    public function provide(string $channelCode): string
+    {
+        /** @var ChannelInterface $channel */
+        $channel = $this->channelRepository->findOneByCode($channelCode);
+
+        return null !== $channel->getColor() ? $channel->getColor() : $this->defaultChannelColor;
```
lchrusciel:
```suggestion
        return $channel->getColor() ?? $this->defaultChannelColor;
```

lchrusciel:
Thank you, Bartosz! :1st_place_medal:

Issue #111, opened 2019-02-22
InvoicePdfFileGenerator should receive invoice & channel models instead of their ids


pamil:
It violates SRP and prevents us from generating a PDF file for an invoice that is not persisted to the database.

PR #125, opened 2019-04-19
Passing invoice entity to invoice pdf file generator

bartoszpietrzak1994:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/111

### DIFF
--- a/spec/CommandHandler/SendInvoiceEmailHandlerSpec.php
+++ b/spec/CommandHandler/SendInvoiceEmailHandlerSpec.php
@@ -5,11 +5,14 @@ declare(strict_types=1);
 namespace spec\Sylius\InvoicingPlugin\CommandHandler;
 
 use PhpSpec\ObjectBehavior;
+use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
+use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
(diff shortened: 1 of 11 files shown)


src/Generator/InvoicePdfFileGenerator.php
```diff
@@ -38,30 +31,22 @@ final class InvoicePdfFileGenerator implements InvoicePdfFileGeneratorInterface
...
-        $this->channelRepository = $channelRepository;
         $this->templatingEngine = $templatingEngine;
         $this->pdfGenerator = $pdfGenerator;
         $this->fileLocator = $fileLocator;
         $this->template = $template;
         $this->invoiceLogoPath = $invoiceLogoPath;
     }
 
-    public function generate(string $invoiceId): InvoicePdf
+    public function generate(InvoiceInterface $invoice, ChannelInterface $channel): InvoicePdf
```
Zales0123: Should we pass a channel to the `generate` function? What if we pass a different channel than the invoice's channel? I think we should still get the channel from the repository basing on data from the invoice, as it would be set even if the invoice would not be persisted to the database.

spec/CommandHandler/SendInvoiceEmailHandlerSpec.php
```diff
@@ -29,7 +31,9 @@ function it_requests_an_email_with_an_invoice_to_be_sent(
         InvoiceEmailSenderInterface $emailSender,
         InvoiceInterface $invoice,
         OrderInterface $order,
-        CustomerInterface $customer
+        CustomerInterface $customer,
+        ChannelInterface $channel,
```
lchrusciel: Is it needed? Same below

Zales0123:
Thank you, Bartosz! :1st_place_medal:

PR #134, opened 2019-06-04
Representative added to invoice shop billing data

bartoszpietrzak1994:


src/Entity/InvoiceShopBillingDataInterface.php
```diff
@@ -29,4 +31,6 @@ public function setStreet(?string $street): void;
     public function setCity(?string $city): void;
 
     public function setPostcode(?string $postcode): void;
+
+    public function setRepresentative(?string $representative): void;
```
GSadee: setter should be next to getter

src/Resources/config/doctrine/InvoiceShopBillingData.orm.xml
```diff
@@ -12,5 +12,6 @@
         <field name="city" nullable="true" />
         <field name="postcode" nullable="true" />
         <field name="countryCode" column="country_code" nullable="true" />
+        <field name="representative" column="representative" nullable="true" />
```
GSadee: column is not needed

src/Entity/InvoiceShopBillingData.php
```diff
@@ -83,4 +91,9 @@ public function setPostcode(?string $postcode): void
     {
         $this->postcode = $postcode;
     }
+
+    public function setRepresentative(?string $representative): void
```
GSadee: as below

GSadee:
Thanks, Bartosz! :tada:

PR #137, opened 2019-06-06
Require Symfony ^4.2 and improve autoloading

pamil:
Symfony 4.1 is only receiving security fixes now.


GSadee:
Thanks, Kamil! :1st_place_medal:

PR #138, opened 2019-06-06
Allow to run PHPUnit without arguments and run Behat with strict mode on Travis CI

pamil:


GSadee:
Thank you, Kamil! :tada:

PR #148, opened 2019-07-18
Fix GenerateInvoicesCommand selects order without number

tom10271:
Supposedly those are cart orders and should not be selected for invoice generation

### DIFF
--- a/src/Cli/GenerateInvoicesCommand.php
+++ b/src/Cli/GenerateInvoicesCommand.php
@@ -31,7 +31,11 @@ final class GenerateInvoicesCommand extends Command
     public function execute(InputInterface $input, OutputInterface $output): void
     {
         /** @var array $orders */
-        $orders = $this->orderRepository->findAll();
+        $orders = $this->orderRepository
+            ->createListQueryBuilder()
+            ->andWhere('o.number IS NOT NULL')
+            ->getQuery()
+            ->getResult();
 
         $this->massInvoicesCreator->__invoke($orders);
 


src/Cli/GenerateInvoicesCommand.php
```diff
@@ -31,7 +31,11 @@ public function __construct(
     public function execute(InputInterface $input, OutputInterface $output): void
     {
         /** @var array $orders */
-        $orders = $this->orderRepository->findAll();
+        $orders = $this->orderRepository
```
lchrusciel: Shouldn't we also skip orders in state `cart`? /cc @Zales0123

tom10271: I thought order in cart will not have order number yet.

lchrusciel: That's correct :D nvm then

tom10271:
There is still one more thing to fix, I called `createQueryBuilder` which lived in EntityRepository, but the type set for `$orderRepository` is `OrderRepositoryInterface`making CI cannot pass.

Which interface or class should we set?

Zales0123:
The best solution would be creating a custom `OrderRepository`, but it requires a custom interface, class, and configuration, which is probably not something we would like to introduce in such a simple fix :/ Maybe you can use one of the existing methods from order repository, like `createListQueryBuilder`, and then apply the not-null-number criteria on the result? It should do the trick with not much work. Thanks for the contribution!

tom10271:
I have updated the commit. Please check again. Thank you everyone

Zales0123:
👍 The last thing that should be done is adding a not-placed order in the background of [this feature file](https://github.com/Sylius/InvoicingPlugin/blob/master/features/managing_invoices/generating_invoices_for_previously_placed_orders.feature) to check does this query for sure works as it should work :) Do you think you would be able to do it, @tom10271?

pamil:
Thank you, @tom10271! :1st_place_medal:

diimpp:
That's a might weird solution. Why not check for Order->paymentState instead? Unpaid/cancelled orders have number, but don't qualify to have invoice.

PR #162, opened 2019-10-17
Require GridBundle v1.7+ with embeddables support

pamil:


GSadee:
Thanks, Kamil! :1st_place_medal:

PR #165, opened 2019-11-23
fix(deps): loosen constraint for symfony/messenger

Gounlaf:
Like https://github.com/Sylius/ShopApiPlugin/pull/602: allow upgrade to symfony 4.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "symfony/framework-bundle": "^4.3",
         "symfony/http-foundation": "^4.3",
         "symfony/http-kernel": "^4.3",
-        "symfony/messenger": "~4.3.0",
+        "symfony/messenger": "^4.3",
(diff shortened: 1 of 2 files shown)


Gounlaf:
> Hello Florian,
> 
> thanks for the contribution! Could you take a look into failing build?

Well, I can fix it if you want, but it's not related to this PR...

It's related to release 4.4.
IMO implemented method (execute) shouldn't have a return type (even if require php version is 7.2) to respect contract/parent method.
Another PR should be done to fix it;

pamil:
There are BC breaks between Messenger 4.3 and 4.4, it might be worth to review them: https://github.com/symfony/symfony/blob/master/UPGRADE-4.4.md#messenger

pamil:
Thanks, Florian! :1st_place_medal:

PR #168, opened 2020-01-03
Define logo-file as parameter

BigBadBassMan:
Hi,
since the logo is passed to the pdf generator separately (outside the template), it should be configurable by users, so it can be replaced with the actual businesses logo.

This PR introduces 2 parameters, one holding the bundle-path to the default logo, and a second one, which can be set by an .env var defaulting to the first path, if not set.
With this 2 step approach, this PR is fully backwards compatible.

In addition, the flex recipe would need an update along with the docs to reflect this changes.


src/Resources/config/services.xml
```diff
@@ -5,6 +5,11 @@
         <import resource="services/*.xml" />
     </imports>
 
+    <parameters>
+        <parameter key="default_logo_file">@SyliusInvoicingPlugin/Resources/assets/sylius-logo.png</parameter>
+        <parameter key="sylius.invoicing.logo_file">%env(default:default_logo_file:SYLIUS_INVOICING_LOGO_FILE)%</parameter>
```
lchrusciel: Can we add `template` to the parameter name?
```suggestion
        <parameter key="sylius.invoicing.template.logo_file">%env(default:default_logo_file:SYLIUS_INVOICING_LOGO_FILE)%</parameter>
```

Also, for now I'm not convinced with `default_logo_file` parameter as a fallback. IMHO, we should go with only one argument.

BigBadBassMan: done.
The thing is, the env-processor accepts only another parameter name for the default clause.
In order to avoid b/c breaks, (because of missing env variable in end user projects), I introduced the ```default_logo_file``` parameter, so that this current behavior is not changed, but can be altered to users needs.

The ```default``` processor was only introduced in Symfony 4.3, so there may be unwanted b/c breaks already - but since Sylius 1.6 is not locked to a specific Symfony 4 version, I felt this was acceptable.

BigBadBassMan:
I have no idea why the travis tests failed, local tests succeeded. Anyway, I'd love to see this merged, if there are no more blockers from your side.

lchrusciel:
Hey @BigBadBassMan, we cannot merge anything until the build is green. Someone has to take a look into the master build and fix it before we will be able to merge this PR.

BigBadBassMan:
> Hey @BigBadBassMan, we cannot merge anything until the build is green. Someone has to take a look into the master build and fix it before we will be able to merge this PR.

I understand, perhaps you could trigger a re-test on travis? 
It seems to me that this may be a transient failure, since I didn't change a single line in the tests, and my local tests (specifically the ecs and phpspec tests) run fine. The failure seems not to be related to my changes.

lchrusciel:
I've restarted the build. However, the master is failing one way or another: https://travis-ci.org/github/Sylius/InvoicingPlugin/builds/676556586

BigBadBassMan:
Hi @lchrusciel , I'm finally back on this, would you care to have another look (and possibly accept ;-) ) my PR?

GSadee (approved):

GSadee:
Thanks, @BigBadBassMan! :1st_place_medal:

PR #169, opened 2020-01-14
Add conflict with GridBundle 1.7.4 and increase phpstan

Tomanhez:


src/Ui/Action/Admin/ResendInvoiceAction.php
```diff
@@ -29,9 +29,10 @@ final class ResendInvoiceAction
     /** @var UrlGeneratorInterface */
     private $urlGenerator;
 
-    /** @var Session */
+    /** @var Session|Session[] */
```
GSadee: I'm not conviced to that, as the `$session` argument has `Session` type. Maybe we should exclude that, wdyt @pamil @Zales0123 @lchrusciel

Zales0123: It's quite strange such a change is required... You're sure it's not working without it 😄 if no, I would vote to exclude this check and open an issue to investigate it properly

Tomanhez: I excluded it: https://github.com/Sylius/InvoicingPlugin/pull/169/files#diff-e4ce1df0d2ca17a9d91544a50ace127aR13-R14

src/Ui/Action/Admin/ResendInvoiceAction.php
```diff
@@ -32,6 +32,7 @@ final class ResendInvoiceAction
     /** @var Session */
     private $session;
 
+    /** @param Session $session */
```
pamil: It looks redundant.

src/Ui/Menu/AdminMenuListener.php
```diff
@@ -13,7 +13,7 @@ public function __invoke(MenuBuilderEvent $event): void
     {
         $menu = $event->getMenu();
 
-        /** @var ItemInterface $salesMenu */
+        /** @var ItemInterface|ItemInterface[] $salesMenu */
```
pamil: Do we really need to add this? What was the error behind it?

Tomanhez: I exclude it. But ItemInterface and Session have IteratorAggregate interface and neon seeing it as iterable object 
Screenshot 2020-02-12 at 11 38 28
Screenshot 2020-02-12 at 11 39 46

diimpp:
What are the general policy on these per variable docblocks? I think main Sylius package as well as Symfony forbids them. Personally I would prefer them to be dropped as well.

PR #172, opened 2020-02-19
Fix invalid configuration framework.messenger config messenger error

JoaquinRuiz:
Invalid configuration for path "framework.messenger": You must specify the "default_bus" if you define more than one bus

### DIFF
--- a/src/Resources/config/config.yml
+++ b/src/Resources/config/config.yml
@@ -13,6 +13,7 @@ knp_snappy:
 
 framework:
     messenger:
+        default_bus: sylius_invoicing_plugin.command_bus
         buses:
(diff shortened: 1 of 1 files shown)


Zales0123:
Hi @JoaquinRuiz! As mentioned in [this issue](https://github.com/Sylius/RefundPlugin/pull/164#issuecomment-554348862), we cannot enforce the default bus, as an application can be using different plugins with symfony messenger. We should probably mention that option in the [installation instruction](https://github.com/Sylius/RefundPlugin#installation).

GSadee:
I believe that `default_bus` is already defined in Sylius, so there is no problem with configuration anymore

PR #175, opened 2020-04-24
Switch to Doctrine/Migrations 3.0-alpha

pamil:
Related to https://github.com/Sylius/Sylius/pull/11389.


src/DependencyInjection/SyliusInvoicingExtension.php
```diff
@@ -20,4 +21,23 @@ public function load(array $config, ContainerBuilder $container): void
...
     }
+
+    public function prepend(ContainerBuilder $container): void
+    {
+        if (!$container->hasExtension('doctrine_migrations') || !$container->hasExtension('sylius_core')) {
+            return;
+        }
+
+        $container->prependExtensionConfig('doctrine_migrations', [
+            'migrations_paths' => [
```
lchrusciel, line 32: Will it be required to do it like that, or can we move these configurations to YAML?

Issue #180, opened 2020-06-06
Saving PDF to disk on generation


Vaalyn:
Hello, I'm currently working on a Sylius shop for the german market and I'm wondering if there are any plans to add saving the generated files to disk. Since the language of the invoice PDF can change when changing the users language (or the template, the translations, etc...) the generator might give you a different invoice file than the user received. In order to correctly archive it according to german law I need this file to be generated and stored somewhere so that I can ensure that the file that the user received has not changed.

I saw this comment on the [InvoiceEmailSender](https://github.com/Sylius/InvoicingPlugin/blob/master/src/Email/InvoiceEmailSender.php#L38) and wanted to ask if it is a conscious decision to not save the file to disk immediately after creation:

> // Since Sylius' Mailer does not support sending attachments which aren't real files
> // we have to simulate the file being on the local filesystem, so that we save the PDF,
> // run the callable and delete it when the callable is finished.

If this is not the case would you mind if I create a PR adding a configurable filesystem where those files will be stored when the `CreateInvoiceOnOrderPlacedListener` executes?

Ideally I would set this up with [oneup/flysystem-bundle](https://packagist.org/packages/oneup/flysystem-bundle) so that it can be configured to use all kinds of storage locations. Since this would add a new dependency to the plugin I would like to ask for your opinion on this ~before working on it~.

Edit: I've gone ahead and added this feature to the plugin and submitted a PR for it.

arjanfrans:
I ran into a similar problem. I added my own customer Mailer Adapater in order to use Symfony Mailer with an async transport. However, I can't send any e-mails because the pdf files are gone before the async message handler can send the e-mail.

michaelKaefer:
We also need this functionality.

GSadee:
I'm closing this issue as mentioned feature has been already merged to plugin: https://github.com/Sylius/InvoicingPlugin/pull/231

PR #181, opened 2020-06-09
Add flysystem to store generated invoice PDFs

Vaalyn:
This PR adds the functionality to automatically save the generated PDF invoice to a filesystem location and serve the generated file to users and admins so that is possible to ensure that everyone has the same file for their archives since it is currently possible that the invoice layout or language changed between the user receiving the file and an admin archiving it for example by  downloading it in the backend.

By default the filesystem location is defined as in memory so that the current behaviour of the plugin is maintained. I've added a section to the README.md with a configuration that changes the behaviour to use a local filesystem with the directory at `%kernel.project_dir%/var/storage/invoice` and an explanation on how to use other adapters.

See #180 for my initial post about this functionality for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,27 @@ The dot's color is dependant on a property defined on Channel entity or, if not 
 
 Like any other parameter, `default_channel_color` can also be overwritten in your `config.yml` file.
 
+## Using different filesystem adapters
+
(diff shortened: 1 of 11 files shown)


michaelKaefer:
We also need this functionality.

GSadee:
I'm closing this PR as mentioned feature has been already merged to plugin: https://github.com/Sylius/InvoicingPlugin/pull/231

PR #186, opened 2020-10-15
[Enhancement] Refactor and deprecate Invoice repository class

Prometee:
The actual `Sylius\InvoicingPlugin\Repository\DoctrineInvoiceRepository` has been deprecated and a new class has been defined in the `src/Doctrine/ORM` folder with a name following the entity name.
The corresponding interface has been also deprecated and the new one is owning the only required method.

I propose this PR because the actual repository was not correctly used : it was an override of `sylius_invoicing_plugin.custom_repository.invoice` instead of a default value in the resource plugin configuration.

PR #189, opened 2020-10-21
[FIX] Unable to load Invoice correctly if the Invoice entity is extended

jbcr:
When I extend the Invoice entity, the doctrine ORM cannot load correctly all related data.

To solve this issue, I have copied in my project the custom repository with this change.


GSadee:
Closing in favour of #186

PR #200, opened 2020-11-24
Force invoice relations

Prometee:
I found a potential issue happening if you have trouble enhancing the `Invoice` entity, the `billingData` and `shopBillingData` could become null because doctrine could lost the cascade all in some particular entity override cases.
This PR add nullable false to those relations to avoid getting null value in the database.

I also remove the `tests/Application/src/Migrations/Version*.php` because they are no longer required thanks to doctrine migration update.

### DIFF
--- a/src/Migrations/Version20201124102609.php
+++ b/src/Migrations/Version20201124102609.php
@@ -9,13 +9,13 @@ use Doctrine\Migrations\AbstractMigration;
 
 final class Version20201124102609 extends AbstractMigration
 {
-    public function up(Schema $schema) : void
+    public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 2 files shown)


t-n-y:
I had the same issue. After overriding the invoice entity, my billing data and shop billing data were null. 

ow did you deal with that ? 

Only solution i found was to create setters in my custom entity than set them in a listener (that i use to set the new field i created on my custom invoice entity)

lchrusciel:
Hey @Prometee, can you give some more hints when this error occurs? All three relations are set through constructor, so I would expect them to not be nullable

Prometee:
Hi @lchrusciel, this was post a while ago, maybe it has been fixed now.
I'll refresh my memory about it and get back to you with the exact issue I had.

GSadee:
I'm closing the PR as there has been no answer with the exact issue for about a year. However, feel free to reopen it if it is still valid, probably with the newly created migration 🖖🏻

PR #204, opened 2021-01-20
Upgrade to Sylius 1.9

Tomanhez:


.github/workflows/build.yml
```diff
@@ -13,19 +13,27 @@ jobs:
...
 
-        name: "PHP ${{ matrix.php }}, MySQL ${{ matrix.mysql }}, Node ${{ matrix.node }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}"
 
         strategy:
             fail-fast: false
             matrix:
                 php: [7.4, 7.3]
+                symfony: [^4.4, ^5.2]
+                sylius: [~1.8.0, ~1.9.0-RC.1]
```
AdamKasp, line 23:
```suggestion
                sylius: [~1.8.0, ~1.9.0]
```

lchrusciel, line 23:
```suggestion
                sylius: [~1.8.0, ~1.9.0]
```

.github/workflows/build.yml
```diff
@@ -13,19 +13,27 @@ jobs:
...
 
                 exclude:
                     -
                         php: 7.3
                         mysql: 8.0
+                    -
+                        sylius: ~1.8.0
+                        symfony: ^5.2
+                    -
+                        sylius: ~1.9.0-RC.1
```
AdamKasp, line 35:
```suggestion
                        sylius: ~1.9.0
```

lchrusciel, line 35:
```suggestion
                        sylius: ~1.9.0
```

.github/workflows/build.yml
AdamKasp: we already published sylius 1.9 so IMO it should be without RC

.github/workflows/build.yml
```diff
@@ -13,19 +13,27 @@ jobs:
...
 
-        name: "PHP ${{ matrix.php }}, MySQL ${{ matrix.mysql }}, Node ${{ matrix.node }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}"
 
         strategy:
             fail-fast: false
             matrix:
                 php: [7.4, 7.3]
+                symfony: [^4.4, ^5.2]
+                sylius: [~1.8.0, ~1.9]
```
GSadee, line 23: I would be for:
```suggestion
                sylius: [~1.8.0, ~1.9.0]
```
to add next minor in clear way to our tests

.github/workflows/build.yml
```diff
@@ -13,19 +13,27 @@ jobs:
...
                 exclude:
                     -
                         php: 7.3
                         mysql: 8.0
+                    -
+                        sylius: ~1.8.0
+                        symfony: ^5.2
+                    -
+                        sylius: ~1.9
+                        php: 7.3
```
GSadee, line 36: Why did you add a conflict for Sylius 1.9 with PHP 7.3?

.github/workflows/build.yml
```diff
@@ -75,6 +83,23 @@ jobs:
                     restore-keys: |
                         ${{ runner.os }}-php-${{ matrix.php }}-composer-
 
+            -
+                name: Install Sylius version
+                run: |
+                    composer require "sylius/sylius:${{ matrix.sylius }}" --no-update
```
GSadee, line 89: I'm not sure if this is necessary as we have `Restrict Sylius version` in line 99, could you check that?

.github/workflows/build.yml
Tomanhez: Was green, but we droped php 7.3 in Sylius 1.9

GSadee: Nope, we will drop PHP 7.3 in next minor version, we still support this version in 1.9: https://github.com/Sylius/Sylius/blob/1.9/composer.json#L22

Tomanhez:
Closed for : https://github.com/Sylius/InvoicingPlugin/pull/212

Issue #206, opened 2021-02-08
Typing problem on LineItem entity


jorisduenas:
Hi, 

I noticed a problem on LineItem objects. In the class we find getId() and id() which return a string like this:

[LineItem.php#L62](https://github.com/Sylius/InvoicingPlugin/blob/6ae2030e5f90a433a0f62268bda371773c4a7d0a/src/Entity/LineItem.php#L62)
```
public function getId(): string
{
    return $this->id();
}

public function id(): string
{
    return $this->id;
}
```

However, this id is configured as integer in the orm:

[LineItem.orm.xml#L5](https://github.com/Sylius/InvoicingPlugin/blob/6ae2030e5f90a433a0f62268bda371773c4a7d0a/src/Resources/config/doctrine/LineItem.orm.xml#L5)
```
<id name="id" column="id" type="integer">
    <generator strategy="AUTO" />
</id>
```

This creates typing errors.

diimpp:
And it should be nullable as well.

GSadee:
I'm closing the issue as it seems to be already resolved

PR #208, opened 2021-02-19
Fix getId/id return type for entities 

diimpp:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/206

### DIFF
--- a/src/Entity/InvoiceShopBillingData.php
+++ b/src/Entity/InvoiceShopBillingData.php
@@ -43,39 +43,14 @@ class InvoiceShopBillingData implements InvoiceShopBillingDataInterface, Resourc
         return $this->company;
     }
 
-    public function getTaxId(): ?string
-    {
-        return $this->taxId;
-    }
-
-    public function getCountryCode(): ?string
-    {
-        return $this->countryCode;
-    }
-
-    public function getStreet(): ?string
-    {
-        return $this->street;
-    }
-
-    public function getCity(): ?string
-    {
-        return $this->city;
-    }
-
-    public function getPostcode(): ?string
-    {
-        return $this->postcode;
-    }
-
-    public function getRepresentative(): ?string
-    {
-        return $this->representative;
-    }
-
     public function setCompany(?string $company): void
     {
         $this->company = $company;
+    }
+
+    public function getTaxId(): ?string
+    {
+        return $this->taxId;
     }
 
     public function setTaxId(?string $taxId): void
@@ -83,9 +58,19 @@ class InvoiceShopBillingData implements InvoiceShopBillingDataInterface, Resourc
         $this->taxId = $taxId;
     }
 
+    public function getCountryCode(): ?string
+    {
+        return $this->countryCode;
+    }
+
     public function setCountryCode(?string $countryCode): void
     {
         $this->countryCode = $countryCode;
+    }
+
+    public function getStreet(): ?string
+    {
+        return $this->street;
     }
 
     public function setStreet(?string $street): void
@@ -93,14 +78,29 @@ class InvoiceShopBillingData implements InvoiceShopBillingDataInterface, Resourc
         $this->street = $street;
     }
 
+    public function getCity(): ?string
+    {
+        return $this->city;
+    }
+
     public function setCity(?string $city): void
     {
         $this->city = $city;
     }
 
+    public function getPostcode(): ?string
+    {
+        return $this->postcode;
+    }
+
     public function setPostcode(?string $postcode): void
     {
         $this->postcode = $postcode;
+    }
+
+    public function getRepresentative(): ?string
+    {
+        return $this->representative;
     }
 
     public function setRepresentative(?string $representative): void
--- a/src/Entity/InvoiceShopBillingData.php
+++ b/src/Entity/InvoiceShopBillingData.php
@@ -9,7 +9,7 @@ use Sylius\Component\Resource\Model\ResourceInterface;
 /** @final */
 class InvoiceShopBillingData implements InvoiceShopBillingDataInterface, ResourceInterface
 {
-    /** @var string */
+    /** @var int|null */
     protected $id;
 
     /** @var string|null */
@@ -33,7 +33,7 @@ class InvoiceShopBillingData implements InvoiceShopBillingDataInterface, Resourc
     /** @var string|null */
     protected $representative;
 
-    public function getId(): string
+    public function getId(): ?int
     {
         return $this->id;
     }
--- a/src/Entity/LineItem.php
+++ b/src/Entity/LineItem.php
@@ -9,7 +9,7 @@ use Sylius\Component\Resource\Model\ResourceInterface;
 /** @final */
 class LineItem implements LineItemInterface, ResourceInterface
 {
-    /** @var string */
+    /** @var int|null */
     protected $id;
 
     /** @var InvoiceInterface */
@@ -59,12 +59,12 @@ class LineItem implements LineItemInterface, ResourceInterface
         $this->variantCode = $variantCode;
     }
 
-    public function getId(): string
+    public function getId(): ?int
     {
         return $this->id();
     }
 
-    public function id(): string
+    public function id(): ?int
     {
         return $this->id;
     }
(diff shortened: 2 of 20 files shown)


src/Resources/config/doctrine/BillingData.orm.xml
```diff
@@ -1,19 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
```
diimpp, line 3: I've set this "header" as it's done at sylius/sylius

src/Entity/InvoiceShopBillingData.php
```diff
@@ -45,64 +44,64 @@ public function getCompany(): ?string
         return $this->company;
     }
 
-    public function getTaxId(): ?string
+    public function setCompany(?string $company): void
```
diimpp, line 47: Methods re-arranged in order of class variables declaration/doctrine entities declaration

diimpp:
@lchrusciel review please

src/Resources/config/doctrine/BillingData.orm.xml
```diff
@@ -1,19 +1,23 @@
...
-        <field name="provinceName" column="province_name" nullable="true" />
+        <field name="firstName" column="first_name" type="string" />
+        <field name="lastName" column="last_name" type="string" />
+        <field name="company" column="company" type="string" nullable="true" />
+        <field name="street" column="street" type="string" />
+        <field name="city" column="city" type="string" />
+        <field name="postcode" column="postcode" type="string" />
+        <field name="countryCode" column="country_code" type="string" />
+        <field name="provinceCode" column="province_code" type="string" nullable="true" />
+        <field name="provinceName" column="province_name" type="string" nullable="true" />
```
GSadee, line 21: Why did you want to normalize mappings by adding these default values that are not necessary? The same applies to changes in other mappings

src/Entity/LineItem.php
```diff
@@ -64,12 +63,12 @@ public function __construct(
         $this->taxRate = $taxRate;
     }
 
-    public function getId()
+    public function getId(): ?int
```
GSadee, line 66: I would be for leaving it as it was before, without return type to keep id more customizable and consistent with all Sylius resources. I know that our mapping is configured for using integers, but it is relatively easy to change the strategy to use for example UUIDs instead of integers.

GSadee:
@diimpp Sorry for that, especially that the PR have been opened for a long time, but I'm closing it as these are the changes that we don't want to have in this plugin. Of course, feel free to share your thoughts about that if you want to.

diimpp:
@GSadee same change to the RefundPlugin was accepted. https://github.com/Sylius/RefundPlugin/pull/199
There are tests and many other fixes besides `getId(): ?int`.

How come you're closing this?

diimpp:
@GSadee don't you see related bug report as well? https://github.com/Sylius/InvoicingPlugin/issues/206

src/Entity/LineItem.php
diimpp: I would agree that `getId(): ?int` is opinionated, but not because somebody might want to make that UUID, but because sylius resource bundle doesn't define type-hint.

Though this is concrete implementation that you're marking *final*. In such case it's totally alright to make it `int` and not to support some doctrine hacks for 3rd party.
image

src/Resources/config/doctrine/BillingData.orm.xml
diimpp: How come column **type** is not a necessary field? Why rely on doctrine to have expected defaults instead of setting correct type explicitly?

There is difference in how source code is written at `App\` level and `Sylius\Sylius` level.
This is as much documentation as correct way to write entity mappings.

Besides, this is the way all sylius/sylius mappings are done.
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Order.orm.xml

diimpp:
I've rebased PR to the current master, I can remove `getId(): ?int` typehint in favor of `getId()`, but it means no `getId(): string` as well.

src/Entity/LineItem.php
diimpp: Did you see, that BillingData id is already `int` in class, but `ShopBillingData` is not? This is inconsistency and my PR is solving this as well.

diimpp: Should `Invoice`'s `protected string $id` became `@mixed` as well and constructor's `string` type-hint removed, because somebody might want to change DB id to int?

diimpp:
@GSadee @Zales0123 @CoderMaggie hey guys, closure of this PR needs to be re discussed.

There are bugs, that this PR closes and very similar PR for RefundPlugin was merged. I've answered above for all raised points.

https://github.com/Sylius/InvoicingPlugin/pull/208

src/Resources/config/doctrine/BillingData.orm.xml
GSadee: The default value of **type** is `string`, so IMO there is no need to add it here, as they weren't defined before, especially in the PR with the title `Fix getId/id return type for entities`, if you would like to change that, it should be done in the separate PR

diimpp: Doctrine mappings and entity fields types are quite related. Which issue will be solved by moving this to separate PR?

Please ask for 3rd opinion within the core team. Personally I don't see what there to discuss, `sylius/sylius` defines explicit doctrine mappings and that's how it should be done.

If you have different opinion on the matter -- please open RFC at `sylius/sylius` to remove string `type` definition from mappings.

GSadee: @diimpp you've opened a PR with fixing getId return types, but you've also refactored other methods and the corresponding mapping, this refactoring should be done in a separate PR.

GSadee: And fyi, I've already asked @Zales0123 and @lchrusciel for other opinions 😃

Zales0123:
Hello @diimpp! I once again, after @GSadee, apologise for such a long response time. Going straight to the problems about the review here.

1. I would definitely bump the idea of extracting some changes from this PR. It's named `Fix getId/id return type for entities` but there are definitely more things done here. You doubted in [this comment](https://github.com/Sylius/InvoicingPlugin/pull/208#discussion_r772313996) that extracting these changes would make any difference, but the conversation around them clearly shows, that the opinions and comments touch different areas of application (e.g. configuration vs implementation) - which can be a clue that these are separate changes and should be processed separately :) 
2. I would resign from changing the return type of `getId()` methods. You're both (@diimpp and @GSadee) right - this is not defined in the ResourceBundle so should not be here as well (first premise) **and because of that** can be configured to be e.g. UUID rather than int and we should not block it (second premise) :dancer:
3. I'm having a big dilemma with default types in mapping configuration 💥 On the one hand, **indeed** in most of the places in Sylius we define them. I don't know is it because when Sylius started to be developed it was not yet clear in Doctrine? Or maybe our approach was different? 🤷‍♂️ No clue. On the other hand, we trust Symfony and Doctrine as our main dependencies and have a great test suite, that would protect us from any differences if those happens :) So I would **personally** be for not defining them... but I have no strong opinion about that. Maybe @lchrusciel has some opinion? 

Cheers and sorry one more time for such a delay in answer 🖖

diimpp:
@Zales0123 Hi, thank you for update.

1. Please feel free to re-use any commits or changes from this PR.
2. You need to decide on the strategy, there are all three options already in the codebase with no type hints, with string and int type hints. There is open bug with overly strict int type hint as well, which is open for more than a year now. (Side point, both of you suggesting somebody might want to change ID field type to string/UUID, but think about this from regular developer view point -- you install bundle from packagist and you're going to change ID mappings of 3rd party code without and public contract or official support? Obviously not, acceptable solution is to introduce `publicId` property with any data type and not to touch original mappings, so this case is not important for consideration)
3. Explicit doctrine mappings are form of documentation as well as already selected approach for sylius/sylius:1. Look at this from end-developer view point, who just does another task for the acme company and doesn't care for doctrine or sylius defaults -- those explicit mappings helps such developer.
I remember there was some weird field on some entity either in this plugin or refund plugin, where I couldn't detect field content from the codebase and mappings were missing. I ended up experimenting on mysql5.7,8 to see which defaults would be real. What about postgresql in this case? There is just too much uncertainty for plugin's end user.

PR #209, opened 2021-02-24
Fix invoice pdf tpl |trans usage to invoice locale instead of curret user locale

diimpp:
Currently pdf is generated in current user locale instead of invoice locale, which will result in incorrect translations in pdf generated from admin panel, if admin user locale doesn't match shop user locale.

Same issue and same solution as for SyliusRefundPlugin.


diimpp:
Solved at https://github.com/Sylius/InvoicingPlugin/pull/251

PR #226, opened 2021-06-11
Relate Invoice with Order

Zales0123:
Similar to https://github.com/Sylius/RefundPlugin/pull/306 and https://github.com/Sylius/RefundPlugin/pull/307.

Should be easier to operate on the Invoice with a proper relation rather than just an order number

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -7,7 +7,6 @@
    `Sylius\InvoicingPlugin\Doctrine\ORM\InvoiceRepositoryInterface::findOneByOrder`.
 1. `Sylius\InvoicingPlugin\Factory\InvoiceFactoryInterface::createForData` takes `OrderInterface $order` as the 3rd argument instead
     of `string $orderNumber`.
-1. 
 
 ### UPGRADE FROM 0.14.0 TO 0.15.0
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,14 @@
+### UPGRADE FROM 0.15.0 TO 0.16.0
+
+1. `orderNumber` field on `Sylius\InvoicingPlugin\Entity\Invoice` has been removed and replaced with relation to `Order` entity.
+1. `Sylius\InvoicingPlugin\Entity\InvoiceInterface::orderNumber` function is left due to easier and smoother upgrades,
+   but is also deprecated and will be removed in the `v1.0.0` release. Use `Sylius\InvoicingPlugin\Entity\InvoiceInterface::order` instead.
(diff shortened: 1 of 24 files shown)


GSadee (commented):
Could you add a note in UPGRADE file?

Zales0123:
@GSadee UPGRADE file updated 🖖

UPGRADE.md
```diff
@@ -1,3 +1,14 @@
+### UPGRADE FROM 0.15.0 TO 0.16.0
+
+1. `orderNumber` field on `Sylius\InvoicingPlugin\Entity\Invoice` has been removed and replaced with relation to `Order` entity.
+1. `Sylius\InvoicingPlugin\Entity\InvoiceInterface::orderNumber` function is left due to easier and smoother upgrades,
+   but is also deprecated and will be removed in the `v1.0.0` release. Use `Sylius\InvoicingPlugin\Entity\InvoiceInterface::order` instead.
+1. `Sylius\InvoicingPlugin\Doctrine\ORM\InvoiceRepositoryInterface::findOneByOrderNumber` method has been replaced by
+   `Sylius\InvoicingPlugin\Doctrine\ORM\InvoiceRepositoryInterface::findOneByOrder`.
+1. `Sylius\InvoicingPlugin\Factory\InvoiceFactoryInterface::createForData` takes `OrderInterface $order` as the 3rd argument instead
+    of `string $orderNumber`.
+1.
```
GSadee, line 10:
```suggestion
```

UPGRADE.md
```diff
@@ -1,3 +1,14 @@
+### UPGRADE FROM 0.15.0 TO 0.16.0
+
+1. `orderNumber` field on `Sylius\InvoicingPlugin\Entity\Invoice` has been removed and replaced with relation to `Order` entity.
+1. `Sylius\InvoicingPlugin\Entity\InvoiceInterface::orderNumber` function is left due to easier and smoother upgrades,
+   but is also deprecated and will be removed in the `v1.0.0` release. Use `Sylius\InvoicingPlugin\Entity\InvoiceInterface::order` instead.
```
GSadee, line 5: Shouldn't we remove that method earlier than v1.0.0?

Zales0123: I chose `v1.0.0` because I had no better idea which version it should be :D alpha? beta? rc? I think we can rethink that, but also we can just change this decision later, so it does not block the merge 💃

GSadee (approved):

GSadee:
Thank you, Mateusz! :tada:

PR #231, opened 2021-07-13
Save invoices on server during generation

Zales0123:
To keep invoices files in exactly the same state (in both template and data) as in the moment of the order being paid, we need to save them on the server, rather than creating them on the fly.

This PR also includes some refactoring and code improvements, to leave the code in a better shape than it was when I started to work with it 💃


src/Provider/InvoiceFileProvider.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->invoicesDirectory = $invoicesDirectory;
+    }
+
+    public function provide(InvoiceInterface $invoice): InvoicePdf
+    {
+        $invoiceFileName = $this->invoiceFileNameGenerator->generateForPdf($invoice);
+        $invoiceFile = $this->filesystem->get($invoiceFileName);
+
+        $invoicePdf = new InvoicePdf($invoiceFileName, $invoiceFile->getContent());
+        $invoicePdf->setFullPath($this->invoicesDirectory . '/' . $invoiceFileName);
```
AdamKasp, line 48: IMHO if we have more than one concatenation we should consider using`sprintf`

tests/Behat/Resources/suites/admin/managing_invoices.yml
```diff
@@ -62,9 +63,57 @@ default:
...
+                - sylius.behat.context.setup.channel
+                - sylius.behat.context.setup.currency
+                - sylius.behat.context.setup.customer
+                - sylius.behat.context.setup.geographical
+                - sylius.behat.context.setup.order
+                - sylius.behat.context.setup.payment
+                - sylius.behat.context.setup.product
+                - sylius.behat.context.setup.product_taxon
+                - sylius.behat.context.setup.promotion
+                - sylius.behat.context.setup.admin_security
```
AdamKasp, line 80: Do we really need all of these contexts?

src/Provider/InvoiceFileProvider.php
```diff
@@ -0,0 +1,52 @@
...
+    ) {
+        $this->invoiceFileNameGenerator = $invoiceFileNameGenerator;
+        $this->filesystem = $filesystem;
+        $this->invoicesDirectory = $invoicesDirectory;
+    }
+
+    public function provide(InvoiceInterface $invoice): InvoicePdf
+    {
+        $invoiceFileName = $this->invoiceFileNameGenerator->generateForPdf($invoice);
+        $invoiceFile = $this->filesystem->get($invoiceFileName);
```
GSadee, line 45: What if there is no invoice saved in the filesystem? Shouldn't it be generated here?

UPGRADE.md
```diff
@@ -1,3 +1,16 @@
+### UPGRADE FROM 0.16.1 TO 0.17.0
+
+Invoices are now saved on the server during their generation (by default, when the order is paid).
+
+1. `Sylius\InvoicingPlugin\Creator\InvoiceCreator` class has 2 more dependencies: `InvoicePdfFileGeneratorInterface $invoicePdfFileGenerator`
```
AdamKasp, line 5: I am wondering if it will more clear if we show all constructor parameters instead of describing what have we changed.

AdamKasp (approved):

GSadee (approved):

src/Provider/InvoiceFileProvider.php
GSadee: And what is more, the current solution is not backwards compatible, as there will be no generated pdfs in the filesystem, so no one will be able to download the previously created invoice after this change.

Zales0123: Agreed 👍 It should probably be generated here, to keep it both simple enough and backward compatible

features/managing_invoices/saving_invoices_on_server_during_generation.feature
```diff
@@ -0,0 +1,23 @@
...
+        And the store allows paying with "Cash on Delivery"
+        And channel "United States" has shop billing data set as "Ragnarok", "1100110011", "Pacific Coast Hwy", "90806" "Los Angeles", "United States"
+
+    @application
+    Scenario: Having invoice saved on the server after the order is placed
+        Given there is a customer "lucy@teamlucifer.com" that placed an order "#00000666"
+        When the customer bought 2 "Angel T-Shirt" products
+        And the customer "Lucifer Morningstar" addressed it to "Seaside Fwy", "90802" "Los Angeles" in the "United States"
+        And for the billing address of "Mazikeen Lilim" in the "Pacific Coast Hwy", "90806" "Los Angeles", "United States"
+        And the customer chose "UPS" shipping method with "Cash on Delivery" payment
```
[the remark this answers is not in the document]
Zales0123: Answer to both this comment and the one above 💃 cc @AdamKasp 

When putting these steps into the background/writing them with the `Given` keyword, we're missing _an action_ in this scenario, which is checked by the _assertion_ in the last step. The reason for that is how our steps for completing the checkout are named. And yes, they definitely seem like setting up the application 🚀 But renaming them a little bit too big work for a small benefit, at least in my opinion. 

So TL;DR - I wanted to emphasize the action that triggers the invoice generation and that's why all of these steps are here and marked as `When` rather than `Given` and/or in the background 🖖

src/Provider/InvoiceFileProvider.php
Zales0123: In general, I agree, but in this specific case it would look like:

```php
sprintf('%s/%s', $this->invoicesDirectory, $invoiceFileName)
```

Which is far less readable than the proposed concatenation in my opinion :D

tests/Behat/Resources/suites/admin/managing_invoices.yml
Zales0123: Probably I could remove some of them 👌

UPGRADE.md
Zales0123: 🤷‍♂️ both works for me 😄

lchrusciel (approved):

spec/Model/InvoicePdfSpec.php
```diff
@@ -31,4 +31,10 @@ public function it_has_content(): void
     {
         $this->content()->shouldReturn('pdf content');
     }
+
+    function it_has_full_path(): void
```
GSadee, line 35:
```suggestion
    function it_has_a_full_path(): void
```

GSadee (approved):

GSadee:
Thank you, Mateusz! :tada:

PR #232, opened 2021-07-13
Do not require visibility in PHPSpec functions

Zales0123:
According to the convention proposed by [PHPSpec itself](http://www.phpspec.net/en/stable/manual/getting-started.html#specifying-behaviour), as well as [convention on Sylius/Sylius repository](https://github.com/Sylius/Sylius/blob/master/ecs.php#L26).


GSadee (approved):

GSadee:
Thank you, Mateusz! :tada:

PR #239, opened 2021-08-03
Add existing condition to change AdminMenu

DennisdeBest:
Hi, 

I have multiple dashboards depending on the user's ROLE.
One of them has the salesMenu removed on my custom AdminMenuListener.

This worked fine before adding the invoice plugin but now that I have it enabled the AdminMenuListener from the bundle runs after my custom one where the sales menu is removed which triggers a 500 error `Call to a member function addChild() on null` adding this condition makes it work corectly for all my dashboards.

### DIFF
--- a/src/Ui/Menu/AdminMenuListener.php
+++ b/src/Ui/Menu/AdminMenuListener.php
@@ -24,11 +24,12 @@ final class AdminMenuListener
 
         /** @var ItemInterface $salesMenu */
         $salesMenu = $menu->getChild('sales');
-
-        $salesMenu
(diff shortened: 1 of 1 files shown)


GSadee:
Hello @DennisdeBest, to be honest, it seems like a problem with your customization, not with the plugin. In this plugin, we are basing on Sylius menu items and we need to add this invoices section. In my opinion, you should change the order of listeners so that your custom listener will be called the last one.

DennisdeBest:
thanks @GSadee, Indeed I don't know why I hadn't thought of this before. Setting my own custom AdminMenuListener priority to -20 fixed the issue.

PR #245, opened 2021-08-10
Make Invoice a sylius resource

arti0090:


src/Resources/config/app/config.yml
```diff
@@ -0,0 +1,5 @@
+sylius_resource:
+    resources:
+        sylius_invoicing.invoice:
+            classes:
+                model: Sylius\InvoicingPlugin\Entity\Invoice
```
GSadee, line 5: `Invoice` is already defined as a resource in configuration tree: https://github.com/Sylius/InvoicingPlugin/blob/master/src/DependencyInjection/Configuration.php#L53

src/Entity/Invoice.php
```diff
@@ -94,66 +95,173 @@ public function __construct(
...
     {
         return $this->number;
     }
 
+    public function getNumber(): string
+    {
+        return $this->number;
+    }
+
+    public function setNumber(string $number): void
```
GSadee, line 125: `Invoice` is an immutable entity, we have a constructor for creating a proper instance, so we shouldn't add set methods

GSadee (changes requested):

PR #255, opened 2021-09-22
Added missing french translations

Snowbaha:


src/Resources/translations/messages.fr.yml
```diff
@@ -11,7 +12,18 @@ sylius_invoicing_plugin:
...
+        net_value: 'Prix TTC'
+        no: 'N°'
         order_number: 'Numéro de commande'
+        payment:
+            paid: 'Payée'
+            yes: 'Oui'
+            no: 'Non'
         resend_invoice: 'Réenvoyer'
         seller: 'Vendeur'
+        tax_amount: 'Montant TVA'
```
Roshyo, line 25: IMO it should rather be `Montant taxes` because taxes can be something else than VAT only.

And then, if your shop contains only VAT as tax, you can override the translation locally

Snowbaha: I was asking myself if it is better to write `TVA` or `Taxe`, but I think you are right.

Roshyo (approved):
Approving, for what it's worth :D

GSadee (approved):

PR #256, opened 2021-10-20
Fix invoice payment state not updating when given order is paid

GregoireDucharme:
Change conditions settings Invoice Payment state.

`$order->getPaymentState()` is returning a string from `OrderPaymentStates`, therefore Invoice status is never set to `'completed'`


GSadee (approved):

GSadee:
Thanks, Gregoire! :1st_place_medal:

Issue #254, opened 2021-09-22
change the .env SYLIUS_INVOICING_LOGO_FILE


Snowbaha:
Hello,

I would like to change the path of the logo but I don't understand how to do. Can you complete the DOC about this?

I saw this code:
```
<parameters>
        <parameter key="default_logo_file">@SyliusInvoicingPlugin/Resources/assets/sylius-logo.png</parameter>
        <parameter key="sylius.invoicing.template.logo_file">%env(default:default_logo_file:SYLIUS_INVOICING_LOGO_FILE)%</parameter>
    </parameters>
```

so I wanted to add something like: 
`SYLIUS_INVOICING_LOGO_FILE='%kernel.project_dir%/assets/images/logo_xs.png'`

but `%kernel.project_dir%` is not working and I don't know how to write the good path...

Snowbaha:
Hello, I am still stuck, I tried lot of path but nothing is working...

- '../assets/images/logo_xs.png'
-  '@App/assets/images/logo_xs.png'
-  '__DIR__/assets/images/logo_xs.png'
-  'assets/images/logo_xs.png'
-  '*/assets/images/logo_xs.png'
- ...

mifoss:
hi

this should work:
```
//config/services.yaml
parameters:
    ...
    sylius.invoicing.template.logo_file: "%kernel.project_dir%/assets/images/logo_xs.png"
```

GSadee:
I'm closing the issue, as it should be resolved by [PR with a fix](https://github.com/Sylius/InvoicingPlugin/pull/263) from @mifoss and a proper [cookbook for customizing logo](https://github.com/Sylius/Sylius/pull/13421). Feel free to reopen this one or create a new one, if you have still an issue with changing the logo path.

PR #263, opened 2021-12-20
resolve SYLIUS_INVOICING_LOGO_FILE

mifoss:
this should solve the problem with unresolved parameters in .env variables  from #254


GSadee (approved):

GSadee:
Thank you, @mifoss! :tada:

PR #265, opened 2022-01-03
move resource config from php to yaml

AdamKasp:
resource configuration should be the same in all plugins, in resource plugin and Plus we have a configuration in YAML so i moved configs from PHP to YAML.


tests/Application/config/packages/sylius_invoicing.yaml
```diff
@@ -1,23 +1,23 @@
 imports:
     - { resource: "@SyliusInvoicingPlugin/Resources/config/config.yml" }
-            
-sylius_invoicing:
+
+sylius_resource:
     resources:
-        billing_data:
+        sylius_invoicing_plugin.billing_data:
```
GSadee, line 6: You should add a note to UPGRADE file about changing the way of overwriting resource classes

GSadee (changes requested):

UPGRADE.md
```diff
@@ -8,6 +8,12 @@ Override `src/Sylius/Bundle/AdminBundle/Resources/views/Common/_channel.html.twi
     - Interface `Sylius\InvoicingPlugin\Provider\ChannelColorProviderInterface` was removed.
     - Class `Sylius\InvoicingPlugin\Twig\ChannelColorExtension` and twig filter `sylius_channel_color` were removed.
     - Parameter `default_channel_color` was removed
+4. Way to configure resources has been changed:
+    ```dif
```
GSadee, line 12:
```suggestion
    ```diff
```

GSadee: And this is not an only change, there is also a need to change the resource name, as for example for invoice:
```diff
-invoice:
+sylius_invoicing_plugin.invoice:
```

GSadee:
I'm closing the PR as we should stay with the configuration in PHP and adjust other plugins to the way proposed by this plugin 🖖🏻

PR #277, opened 2022-04-22
[DependencyInjection] Move registering resources to prepend method

GSadee:
Before this change, resources and their parameters were loaded too late and they cannot be used properly for example in a grid or state machine configuration.


AdamKasp (approved):

coldic3 (approved):

AdamKasp:
Thank you, Grzegorz! :tada:

PR #278, opened 2022-04-25
Optional parameter to turn off the PDF generation

GSadee:
I'm introducing the new parameter to turn off the PDF generation:
```
sylius_invoicing:
    pdf_generator:
        enabled: false
```
The main reason for adding this parameter is to not force `wkhtmtopdf` to be installed.


.github/workflows/build.yml
```diff
@@ -155,3 +155,160 @@ jobs:
...
+                    mysql version: "${{ matrix.mysql }}"
+                    mysql root password: "root"
+
+            -
+                name: Remove wkhtmltopdf binary
+                run: rm tests/Application/etc/wkhtmltopdf
+
+            -
+                name: Disable PDF generation
+                run: mv tests/Application/etc/sylius_invoicing_pdf_generation_disabled.yaml tests/Application/config/packages
```
lchrusciel, line 213: if only these two steps are different perhaps we can plug this build again to the the main one with included following values:
```
                php: "8.1"
                symfony: "^5.4"
                sylius: "~1.11.0"
                node: "14.x"
                mysql: "8.0"
                pdf: false
```

and the add 
```suggestion
            -
                name: Remove wkhtmltopdf binary
                if: matrix.pdf != ''
                run: rm tests/Application/etc/wkhtmltopdf

            -
                name: Disable PDF generation
                if: matrix.pdf != ''
                run: mv tests/Application/etc/sylius_invoicing_pdf_generation_disabled.yaml tests/Application/config/packages
```

lchrusciel (approved):

Issue #281, opened 2022-05-25
OrderPlacedProducer is not connected to the state machine


ehibes:
Hi, 
I want to change the time when the invoice is created. I would like it to be connected to the transition named `complete` on `sylius_payment`.
I see that the `OrderPlacedProducer` is not connected to the state machine but listens for doctrine events (postPersist & postUpdate).
Can I try to rewrite the call to the `OrderPlacedProducer` by connecting it to the state machine ?
This could solve the following issues : #174 #235

Prometee:
Hello @ehibes I would also love to see this be rewritten because it create some issues while creating orders through BeHat or fully coded orders by hand.

GSadee:
Hi, very late reply from my side, sorry for that 😓 

I tried to recall the reasons why we decided to base this logic on doctrine events (I also consulted it quickly with @Zales0123) instead of the states machine. In my opinion, the main thing could be the fact that the transaction should be completed before generating the invoice, but based on the `postPersist` and `postUpdate` events, the behavior would be the same as using a state machine callbacks, on my current knowledge.

So if the refactor the implementation to link it to the state machine allows for easier customization, and also it seems to be more consistent with Sylius itself, I think we could change that before releasing a stable plugin version. **However, the moment of generating the invoice should stay the same, as it is when the invoice should be generated for the majority of shops.**

PR #283, opened 2022-06-15
Link order generation to state machine

### LINKED ISSUE #174: Change behavior for the bill generation
A bill is generated when the `sylius_payment` flow is complete. This is something good for business to business but not for business to customer because a "complete" `sylius_payment` is not always paid and an unpaid order should not have a generated bill.

That's why I think we should change this behavior.

## Steps

Current configuration must be updated :

- The `sylius_invoicing_plugin_payment_complete_producer` callback should be in the `sylius_order` state machine, on the `fulfill` transition.

```
winzou_state_machine:
    sylius_order:
        callbacks:
            after:
                sylius_invoicing_plugin_payment_complete_producer:
                    on: ['fulfill']
                    do: ['@sylius_invoicing_plugin.event_producer.order_payment_paid', '__invoke']
                    args: ['object']
```

`Sylius\InvoicingPlugin\EventProducer\OrderPaymentPaidProducer` should be updated because there is no more a PaymentInterface but an OrderInterface. Class should be renamed `OrderFulfilledProducer`.

`Sylius\InvoicingPlugin\EventProducer\OrderPlacedProducer` should be updated because the event is no more dispatched with a condition on `checkoutState` (but with `paymentState` and `OrderPaymentStates::STATE_PAID`). Class should be renamed `OrderPaidProducer`.

Related code should be renamed.

## After changes

An unpaid order will not have any bill until the `OrderPaymentStates::STATE_PAID` is dispatched. Then the bill will be available for customer. Because the payment is paid, the owner can print the bill and prepare his order. Finally the order will be "fulfilled" when the shipment is marked as "shipped" and the bill will be sent to the customer.

### LINKED ISSUE #235: Invoice not generated at the right state
The invoice is generated when the order checkout state is completed even if the payment is not paid.

ehibes:
Moving order generation to state machine transition from doctrine persist event. #281
Solve #174 #235 adding some state configuration : 
```yaml
winzou_state_machine:
    sylius_payment:
        callbacks:
            after:
                sylius_invoicing_plugin_order_placed_producer:
                    on: ['complete']
                    do: ['@sylius_invoicing_plugin.event_producer.order_placed', '__invoke']
                    args: ['object.getOrder()']
                sylius_invoicing_plugin_payment_complete_producer:
                    disabled: true
    sylius_order:
        callbacks:
            after:
                sylius_invoicing_plugin_order_placed_producer:
                    disabled: true
```

I had to use the identifier instead of the number, because the order number is not yet persisted to the entity at the time of the transition.


src/Resources/config/config.yml
```diff
@@ -25,6 +25,13 @@ winzou_state_machine:
                     on: ['complete']
                     do: ['@sylius_invoicing_plugin.event_producer.order_payment_paid', '__invoke']
                     args: ['object']
+    sylius_order:
+        callbacks:
+            after:
+                sylius_invoicing_plugin_order_placed_producer:
+                    on: ['create']
+                    do: ['@sylius_invoicing_plugin.event_producer.order_placed', '__invoke']
+                    args: ['object']
```
Prometee, line 34: Reading what the `OrderPlacedProducer` :

- onPostPersist : when `order.checkout_state === 'completed'` then dispatch creation of `sylius_invoicing_plugin_invoice` new item
- onPostUpdate : when the new value of `order.checkout_state === 'completed'` then dispatch creation of `sylius_invoicing_plugin_invoice` new item

Don't you think the new callback should be set on the state machine named : `sylius_order_checkout` on `['complete']` ?

Prometee (commented):
Hello @ehibes !

Why do you change all the `orderNumber` references to `id` ? 

Changing the moment where the invoice item is created could help.

ehibes:
> Hello @ehibes !
> 
> Why do you change all the `orderNumber` references to `id` ?
> 
> Changing the moment where the invoice item is created could help.

I did it because `orderNumber` was not yet generated.

src/Resources/config/config.yml
ehibes: I'll try this, and check if `order_number` is available at this time.
All behat tests passed for me... not for you ?

Prometee: @ehibes reproducing the change on my project I see where you had issue. The command is handle by the `InvoiceCreator` but the repository method used to retrieve the `Order` is `findOneByNumber`, which is a real repository method retrieving an `Order` with a `state` different from `cart`.

A way to solve this is to use this method instead : https://github.com/Sylius/InvoicingPlugin/blob/master/src/Doctrine/ORM/InvoiceRepositoryInterface.php#L24
This method can return multiple `Invoice`s so a `foreach` will be required then.
Finally, the same changes must be done inside `SendInvoiceEmailHandler`.

Prometee: EDIT: Better use `$order = $this->invoiceRepository->findBy(['number' => $orderNumber]);` instead.

src/Creator/InvoiceCreator.php
```diff
@@ -35,16 +35,16 @@ public function __construct(
     ) {
     }
 
-    public function __invoke(string $orderNumber, \DateTimeInterface $dateTime): void
+    public function __invoke(int $orderId, \DateTimeInterface $dateTime): void
     {
         /** @var OrderInterface $order */
-        $order = $this->orderRepository->findOneByNumber($orderNumber);
+        $order = $this->orderRepository->find($orderId);
```
Prometee, line 41:
```suggestion
        /** @var OrderInterface|null $order */
        $order = $this->orderRepository->findOneBy(['number' => $orderNumber]);
        Assert::notNull($order, sprintf('Order with number "%s" does not exist.', $orderNumber));
```

Prometee: The Behat test is maybe missing something, in my current projet I have to prepare my order doing this :

```php
$stateMachine = $this->stateMachineFactory->get($order, OrderTransitions::GRAPH);

$stateMachine->apply(OrderTransitions::TRANSITION_CREATE);
```

save it to the database and complete the order (`OrderCheckoutTransitions::TRANSITION_COMPLETE`).

ehibes:
I made changes according to you reviews. PhpSpec & behat tests passed. I don't know how to add your behat scenario...

spec/Creator/MassInvoicesCreatorSpec.php
```diff
@@ -34,19 +34,19 @@ function it_requests_invoices_creation_for_multiple_orders(
         OrderInterface $secondOrder,
         OrderInterface $thirdOrder
     ): void {
-        $firstOrder->getNumber()->willReturn('0000001');
-        $secondOrder->getNumber()->willReturn('0000002');
-        $thirdOrder->getNumber()->willReturn('0000003');
+        $firstOrder->getNumber()->willReturn('000001');
+        $secondOrder->getNumber()->willReturn('000002');
+        $thirdOrder->getNumber()->willReturn('000003');
```
Prometee, line 39: To ease the reviewing of your PR by Sylius core team members, can you revert it by adding the missing zero ?

spec/Creator/MassInvoicesCreatorSpec.php
```diff
@@ -34,19 +34,19 @@ function it_requests_invoices_creation_for_multiple_orders(
...
         $thirdInvoiceDateTime = new \DateTimeImmutable('2019-02-25');
 
         $dateTimeProvider->__invoke()->willReturn($firstInvoiceDateTime, $secondInvoiceDateTime, $thirdInvoiceDateTime);
 
-        $invoiceCreator->__invoke('0000001', $firstInvoiceDateTime)->shouldBeCalled();
-        $invoiceCreator->__invoke('0000002', $secondInvoiceDateTime)->shouldBeCalled();
-        $invoiceCreator->__invoke('0000003', $thirdInvoiceDateTime)->shouldBeCalled();
+        $invoiceCreator->__invoke('000001', $firstInvoiceDateTime)->shouldBeCalled();
+        $invoiceCreator->__invoke('000002', $secondInvoiceDateTime)->shouldBeCalled();
+        $invoiceCreator->__invoke('000003', $thirdInvoiceDateTime)->shouldBeCalled();
```
Prometee, line 49: Same here

spec/EventListener/CreateInvoiceOnOrderPlacedListenerSpec.php
```diff
@@ -28,8 +28,8 @@ function it_requests_invoice_creation(InvoiceCreatorInterface $invoiceCreator):
     {
         $issuedAt = new \DateTimeImmutable();
 
-        $invoiceCreator->__invoke('0000001', $issuedAt)->shouldBeCalled();
+        $invoiceCreator->__invoke('000001', $issuedAt)->shouldBeCalled();
 
-        $this(new OrderPlaced('0000001', $issuedAt));
+        $this(new OrderPlaced('000001', $issuedAt));
```
Prometee, line 33: And here 😉

src/Creator/InvoiceCreator.php
```diff
@@ -38,7 +39,8 @@ public function __construct(
     public function __invoke(string $orderNumber, \DateTimeInterface $dateTime): void
     {
         /** @var OrderInterface $order */
```
Prometee, line 41: This phpdoc is missing the possibility of `null`

Prometee:
> I made changes according to you reviews. PhpSpec & behat tests passed. I don't know how to add your behat scenario...

Perfect, thank you for your work, I will ping the devs about it to start approving it 😉

Prometee (approved):

Prometee:
Yes it will avoid getting the issue again indeed 👌

.php-version
```diff
@@ -0,0 +1 @@
+8.0
```
GSadee, line 1: This file should be removed

src/Creator/InvoiceCreator.php
```diff
@@ -37,8 +38,9 @@ public function __construct(
 
     public function __invoke(string $orderNumber, \DateTimeInterface $dateTime): void
     {
-        /** @var OrderInterface $order */
-        $order = $this->orderRepository->findOneByNumber($orderNumber);
+        /** @var OrderInterface|null $order */
+        $order = $this->orderRepository->findOneBy(['number' => $orderNumber]);
```
GSadee, line 42: Why did you change that in this PR?

src/Resources/config/config.yml
```diff
@@ -25,6 +25,13 @@ winzou_state_machine:
                     on: ['complete']
                     do: ['@sylius_invoicing_plugin.event_producer.order_payment_paid', '__invoke']
                     args: ['object']
+    sylius_order_checkout:
+        callbacks:
+            after:
+                sylius_invoicing_plugin_order_placed_producer:
+                    on: ['complete']
+                    do: ['@sylius_invoicing_plugin.event_producer.order_placed', '__invoke']
+                    args: ['object']
```
GSadee, line 34: I wonder if it wouldn't be better to put this logic on the callback after `create` transition on `sylius_order` state machine 🤔

src/Creator/InvoiceCreator.php
ehibes: Because `findOneByNumber` refers to OrderRepository method, which verify order is not on cart state. So we have a null value using repository method.
```php
    public function findOneByNumber(string $number): ?OrderInterface
    {
        return $this->createQueryBuilder('o')
            ->andWhere('o.state != :state')
            ->andWhere('o.number = :number')
            ->setParameter('state', OrderInterface::STATE_CART)
            ->setParameter('number', $number)
            ->getQuery()
            ->getOneOrNullResult()
        ;
    }
```

src/Resources/config/config.yml
ehibes: ping @Prometee

src/Creator/InvoiceCreator.php
GSadee: So what is the reason the order is in cart state? Because in the moment of creating an invoice the checkout should be completed and the order shouldn't be in the cart state anymore. I guess, but I'm not sure, there is a difference with the moment of dispatching the event between state machine callback and the doctrine events.

src/Resources/config/config.yml
GSadee: I'm not sure if I correctly understand you, is there a problem with moving the callback to `sylius_order` state machine, in this implementation or in both cases? What is more, as you probably suggested changing the moment of generating an invoice PDF, IMO it is not possible, it should be generated during invoice generation after the order is placed, not payment paid.

src/Creator/InvoiceCreator.php
Prometee: At this moment the order `state` is `new` instead of `cart` in the database.

src/Resources/config/config.yml
Prometee: During the `@sylius_invoicing_plugin.event_producer.order_placed'->__invoke()` the invoice creation is dispatched and cause the error I mentioned. To solve this error we can :

1.  remove the dispatch and simply call the service creating the `Invoice` with a new method signature for the class 
    `Sylius\InvoicingPlugin\Creator\InvoiceCreator`

    From :
    ```php
    public function __invoke(string $orderNumber, \DateTimeInterface $dateTime): void
    ```

    To:
    ```php
    public function __invoke(OrderInterface $order, \DateTimeInterface $dateTime): void
    // Maybe the argument `$dateTime` can be removed and we can use `$order->getCheckoutCompletedAt()` instead
    ```
2. Move the  PDF creation to another moment like when the PDF is sent `payment_state === 'paid'`

Prometee: From my point of view the `Invoice` data stored in the database are the immutable data, the PDF is a view of this data.

David-Crty:
I love this feature. 
Can I help with anything on those falling checks? Let me know if I can.

Rafikooo:
Hello @ehibes,
could you rebase your PR so we can process it again?
Best, Rafał

ehibes:
I don’t think that’s a good choice. After some thought, generating the invoice after the state change and before the Doctrine flush isn’t a good idea.
I think it’s better to use Doctrine events, but it would be nice to easily configure whether or not we wait for the order payment before generating the invoice.

Issue #273, opened 2022-03-30
catalog unit price


laurentdabbb:
Hi, 

I try to adapt invoices to comply with French rules : https://www.economie.gouv.fr/entreprises/factures-mentions-obligatoires (in french).

According to French rules, an invoice must include the catalog unit price : the price of an item before discount application.
Unless I make a mistake, I don't believe that's currently possible.

Within the scope of immutable data, we can't access to the taxRate, so we don't know if the orderItem unitPrice is a net price or a gross price, because the value of "included in price" option is unknown in this context...

I also thought added the amount ot total adjustments to the gross price. But this calculation is not perfect for a free item : the amount of adjustments can be greater than unit price...

The only way I see to get the information is to add the net unit price on the orderItem and to calculate this net unit price in the OrderPricesRecalculator of the OrderProcessor.
Next during invoice generation I store this netUnitPrice on the LineItem entity.

What do you think about storing net unit price on the orderItem and LineItem in a next release ?
If not, how could we recover the value of the net catalog unit price (within the scope of immutable data) ?

Thx

vvasiloi:
You're right, currently the line item's unit price, subtotal and total, all include discounts. Discounts aren't handled in any way.

GSadee:
Hi @laurentdabbb, if I think correctly, the `unitPrice` field on invoice is always a net price as it is [calculated](https://github.com/Sylius/InvoicingPlugin/blob/master/src/Converter/OrderItemUnitsToLineItemsConverter.php#L57) as the difference between gross value and tax amount.

And as for the original unit price, it is already stored on [order item](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Order/Model/OrderItem.php#L39) entity, but not used here in the plugin for line items. And I'm not sure if it will be from the business point of view, wdyt @CoderMaggie? But it is something that you can already change without having to customize order processing.

laurentdabbb:
Hi @GSadee,

Thank you and sorry for the delay in replying, I hadn't seen your reply.

I agree with you, but I'm not talking about the same price ;-) The net price calculated and displayed on the invoice is OK. But it's a net price including discounts.

I want to add a column on invoices to comply with french rules. I had to display the **net sale price** : the net price before discount application. But I can't with data stored on lineItem or orderItem. 
[...]

PR #287, opened 2022-08-16
Unit net price and discounted unit net price on the invoice

Rafikooo:
Fixes https://github.com/Sylius/InvoicingPlugin/issues/273

### DIFF
--- a/src/Converter/ShippingAdjustmentsToLineItemsConverter.php
+++ b/src/Converter/ShippingAdjustmentsToLineItemsConverter.php
@@ -66,6 +66,7 @@ final class ShippingAdjustmentsToLineItemsConverter implements LineItemsConverte
             1,
             $netValue,
             $netValue,
+            $netValue,
             $taxAmount,
             $grossValue,
             null,
--- a/src/Converter/ShippingAdjustmentsToLineItemsConverter.php
+++ b/src/Converter/ShippingAdjustmentsToLineItemsConverter.php
@@ -43,14 +43,16 @@ final class ShippingAdjustmentsToLineItemsConverter implements LineItemsConverte
 
         /** @var AdjustmentInterface $shippingAdjustment */
         foreach ($order->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT) as $shippingAdjustment) {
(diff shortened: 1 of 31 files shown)


AdamKasp (approved):

src/Migrations/Version20220811054008.php
```diff
@@ -0,0 +1,38 @@
...
+final class Version20220811054008 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'original_unit_net_price field added';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE sylius_invoicing_plugin_line_item ADD discounted_unit_net_price INT DEFAULT NULL');
```
lchrusciel, line 31: Shouldn't we add checks similar to Sylius? https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMigration.php#L23

src/Converter/ShippingAdjustmentsToLineItemsConverter.php
```diff
@@ -59,13 +61,15 @@ private function convertShippingAdjustmentToLineItem(AdjustmentInterface $shippi
         $grossValue = $shipment->getAdjustmentsTotal();
         $taxAdjustment = $this->getShipmentTaxAdjustment($shipment);
         $taxAmount = $taxAdjustment !== null ? $taxAdjustment->getAmount() : 0;
-        $netValue = $grossValue - $taxAmount;
+        $unitNetPrice = $grossValue - $taxAmount - $shipment->getAdjustmentsTotal(AdjustmentInterface::ORDER_SHIPPING_PROMOTION_ADJUSTMENT);
```
lchrusciel, line 64: I have the feeling, that it is small de-cohesion, that in the units to line items converter we have dedicated service for net price calculation and here we don't

lchrusciel (approved):

src/Converter/ShippingAdjustmentsToLineItemsConverter.php
Zales0123: Indeed 👍 we could probably refactor (could be done in a separate PR IMO) 🖖

PR #303, opened 2022-12-14
[Maintenance] PHP Attributes introduced

Rafikooo:
Since we do not use PHP 7.4 anymore, it's preferable to support both annotations and attributes with the default annotations setting 🎉


jakubtobiasz (approved):

jakubtobiasz:
Thanks, Rafał! :1st_place_medal:

PR #321, opened 2024-10-01
[symfony/workflow] Makes workflow event listener loading

Prometee:
The event listener for `symfony/workflow` can be loaded even if `winzou/state-machine` is still the default state machine.

Before this PR :

image


diimpp (approved):

GSadee (approved):

GSadee:
Thank you, Francis! :1st_place_medal:

PR #373, opened 2025-06-04
Add sylius/test-application

Wojdylak:


.github/workflows/build.yaml
```diff
@@ -29,21 +29,22 @@ jobs:
...
+#                include:
+#                    -
+#                        php: "8.3"
+#                        symfony: "^7.1"
+#                        sylius: "~2.0.0"
+#                        node: "20.x"
+#                        database: "mysql"
+#                        mysql: "8.4"
+#                        wkhtmltopdf: "0.12.6.1-2"
+#                        state_machine_adapter: "winzou_state_machine"
```
GSadee: Why? As I am guessing, the possibility of conditionally adding bundles in the test application is missing and this is a problem? It would be good to improve the test application for this purpose, possibly here on the build to make it conditionally and set a different env if winzou is selected 🤔

.gitignore
```diff
@@ -10,3 +10,6 @@
 
 /tests/Application/yarn.lock
 /tests/Application/private/invoices/*
```
GSadee: It can be removed now

composer.json
```diff
@@ -10,7 +10,7 @@
         "ramsey/uuid": "^4.7",
         "sylius/grid-bundle": "^1.13",
         "sylius/resource-bundle": "^1.12",
-        "sylius/sylius": "~2.0.0",
+        "sylius/sylius": "^2.0.0",
```
GSadee:
```suggestion
        "sylius/sylius": "^2.0,
```

GSadee (approved):

PR #382, opened 2025-07-30
Proposal to generate an invoice with Gotenberg

### LINKED ISSUE #376: Move away from wkhtmltopdf for pdf generation
wkhtmltopdf has been unmaintained for years and relies on Qt 5 libraries, which lost even commercial extended support in May 2025. Its dependencies are increasingly vulnerable, and no longer receive security patches. Additionally, it's becoming more difficult to find modern hosting providers that still offer wkhtmltopdf. Some are even actively removing it from their servers. To ensure that the plugin can still be used in the near future, it would be wise to consider replacing wkhtmltopdf with a maintained alternative for PDF generation.

Related: #318 #324

StevenRenaux:
Related to https://github.com/Sylius/InvoicingPlugin/issues/376

A simple implementation for GotenbergBundle, potentially replacing wkhtmltopdf.

While integration with GotenbergBundle is possible and desirable, further testing and implementation will be required for a smooth transition of all current use cases.

Rendering by GotenbergBundle
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
         "php": "^8.2",
         "knplabs/knp-snappy-bundle": "^1.10",
         "ramsey/uuid": "^4.7",
+        "sensiolabs/gotenberg-bundle": "^1.0",
         "sylius/grid-bundle": "^1.13",
         "sylius/resource-bundle": "^1.12",
         "sylius/sylius": "^2.0",
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -173,6 +173,7 @@ final class Configuration implements ConfigurationInterface
                     ->children()
                         ->booleanNode('enabled')->defaultTrue()
                     ->end()
+                    ->booleanNode('generate_with_gotenberg')->defaultFalse()
                 ->end()
             ->end()
         ;
--- a/src/DependencyInjection/SyliusInvoicingExtension.php
+++ b/src/DependencyInjection/SyliusInvoicingExtension.php
@@ -15,11 +15,14 @@ namespace Sylius\InvoicingPlugin\DependencyInjection;
 
 use Sylius\Bundle\CoreBundle\DependencyInjection\PrependDoctrineMigrationsTrait;
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
+use Sylius\InvoicingPlugin\Generator\TwigToGotenbergPdfGenerator;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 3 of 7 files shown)


templates/shared/download/pdf_layout.html.twig
```diff
@@ -113,7 +113,7 @@
     <table class="layout">
         <tr>
             <td>
-                <img width="160" src="{{ logoPath }}">
+                <img width="160" src="{{ generateWithGotenberg ? gotenberg_asset(logoPath) : logoPath }}">
```
Neirda24, line 116: inject the config key `$config['pdf_generator']['generate_with_gotenberg']` to wrap this logic in an if maybe ?

composer.json
```diff
@@ -8,6 +8,7 @@
         "php": "^8.2",
         "knplabs/knp-snappy-bundle": "^1.10",
         "ramsey/uuid": "^4.7",
+        "sensiolabs/gotenberg-bundle": "^1.0",
```
Neirda24, line 11: suggest / require-dev ?

src/DependencyInjection/SyliusInvoicingExtension.php
```diff
@@ -35,6 +38,15 @@ public function load(array $configs, ContainerBuilder $container): void
 
         $config = $this->processConfiguration($configuration, $configs);
         $container->setParameter('sylius_invoicing.pdf_generator.allowed_files', $config['pdf_generator']['allowed_files']);
+        $container->setParameter('sylius_invoicing.pdf_generator.generate_with_gotenberg', $config['pdf_generator']['generate_with_gotenberg']);
+
+        if ($config['pdf_generator']['generate_with_gotenberg'] === true) {
+            $definition = new Definition(TwigToGotenbergPdfGenerator::class);
+            $definition->addArgument(new Reference('sensiolabs_gotenberg'));
```
Jean-Beru, line 45: We should check if the bundle is installed and throw an exception if necessary. 
IMO the bundle should be optional.

src/DependencyInjection/SyliusInvoicingExtension.php
```diff
@@ -35,6 +38,15 @@ public function load(array $configs, ContainerBuilder $container): void
...
+            $definition->setDecoratedService('sylius_invoicing.generator.twig_to_pdf');
```
Jean-Beru, line 46: Is decoration needed since we just replace the service?

src/DependencyInjection/Configuration.php
```diff
@@ -173,6 +173,7 @@ private function addPdfGeneratorSection(ArrayNodeDefinition $node): void
                     ->children()
                         ->booleanNode('enabled')->defaultTrue()
                     ->end()
+                    ->booleanNode('generate_with_gotenberg')->defaultFalse()
```
Jean-Beru, line 176: What about adding a `service` property to set the PDF generator to use? It could be `Wkhtmltopdf` by default to avoid BC break and `Gotenberg` if the bundle is installed.

PR #383, opened 2025-08-07
Multiple invoice table row fix in shop order page

acirtautas:
If order has more than one invoice, 
rendering in shop order page is broken as all rows are placed inside one `<tr>` element instead of individual ones.


Rafikooo (approved):

Rafikooo:
Thank you, @acirtautas.  
In the future, it would be great if you could attach a visual comparison of the things you fix, if applicable 💃

PR #387, opened 2025-09-23
Add invoices column to orders grid

TheMilek:
New clickable download buttons on orders grid

image

### DIFF
--- a/templates/shop/account/order/grid/field/invoices.html.twig
+++ b/templates/shop/account/order/grid/field/invoices.html.twig
@@ -1,0 +1,17 @@
+{% set invoices = invoices_for_order(data) %}
+{% if invoices is empty %}
+    <span class="text-muted">—</span>
+{% else %}
+    {% set pdfEnabled = invoices_pdf_enabled() %}
+    {% for invoice in invoices %}
+        {% if pdfEnabled %}
+            <div>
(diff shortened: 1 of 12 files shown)


templates/shop/account/order/grid/field/invoices.html.twig
```diff
@@ -0,0 +1,17 @@
+{% set invoices = invoices_for_order(data) %}
+{% if invoices is empty %}
+    <span class="text-muted">—</span>
+{% else %}
+    {% set pdfEnabled = invoices_pdf_enabled() %}
```
mpysiak:
```suggestion
    {% set pdf_enabled = invoices_pdf_enabled() %}
```

tests/Behat/Context/Ui/Shop/AccountContext.php
```diff
@@ -0,0 +1,41 @@
...
+use Webmozart\Assert\Assert;
+
+final class AccountContext implements Context
+{
+    public function __construct(private IndexPageInterface $orderIndexPage)
+    {
+    }
+
+    /**
+     * @Then I should be able to download an invoice from my orders list
```
mpysiak: you can use new php attributes here

mpysiak (approved):

PR #397, opened 2025-10-22
Isolate plugin messaging: Introduce dedicated event & command buses for InvoicingPlugin

### LINKED ISSUE #157: Shouldn't _this plugin's_ MessageHandlers be configured to only process messages from _this plugin's_ message buses ?
I am developing a plugin for sylius, and, as with this plugin, I am using messenger.

When debugging my code with `php bin/console debug:messenger` I can see that this plugins message handlers are listening to all buses.

While this is not an urgent issue, and may not even produce any errors in most cases, surely, for maximum interoperability with other plugins/bundles, unless there is a specific reason why these handlers **need** to listen on other buses, they should probably be restricted to the buses defined in this bundle.

tomkalon:
| Q               | A
|-----------------|-----
| Branch?         | 3.0
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | https://github.com/Sylius/InvoicingPlugin/issues/157
| License         | MIT

Isolate plugin messaging: Introduce dedicated event & command buses for InvoicingPlugin

### DIFF
--- a/CHANGELOG-3.0.md
+++ b/CHANGELOG-3.0.md
@@ -1,0 +1,5 @@
+# CHANGELOG
+
+### v3.0.0 (2025-10-22)
+
+- [#397](https://github.com/Sylius/InvoicingPlugin/pull/397) Isolate plugin messaging: Introduce dedicated event & command buses for InvoicingPlugin ([@tomkalon](https://github.com/tomkalon))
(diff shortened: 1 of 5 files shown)


NoResponseMate (approved):
But change the base to 3.0 instead of 2.1

tomkalon:
@NoResponseMate 3.0 isn’t out yet. Shall we put it in the minor section? For example 2.3?

